PR #3, opened 2021-09-15
Add missing types

shjortConcordium:
## Purpose

Add missing fields to types from node.

## Changes

Adds `AccountBakerDetails` to `AccountInfo`.
Split `EventResult` to in `SuccessfulEventResult` and `RejectedEventResult`. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -92,10 +92,69 @@ export interface TransferredEvent {
     from: AddressAccount;
 }
 
+/**
+ * An enum containing all the possible reject reasons that can be
+ * received from a node as a response to a transaction submission.
+ *
+ * This should be kept in sync with the list of reject reasons
+ * found here: https://github.com/Concordium/concordium-base/blob/main/haskell-src/Concordium/Types/Execution.hs
+ */
+export enum RejectReasonTag {
+    ModuleNotWF = 'ModuleNotWF',
+    ModuleHashAlreadyExists = 'ModuleHashAlreadyExists',
+    InvalidAccountReference = 'InvalidAccountReference',
+    InvalidInitMethod = 'InvalidInitMethod',
+    InvalidReceiveMethod = 'InvalidReceiveMethod',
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -511,6 +515,27 @@ export interface InitialAccountCredential {
...
+          epoch: number;
+      }
+    | {
+          change: 'RemoveBaker';
+          epoch: number;
+      };
+
+export interface AccountBakerDetails {
+    restakeEarnings: boolean;
+    bakerId: number;
```
orhoj, line 531: This is a `Word64` so it cannot be contained in a number.

src/types.ts
```diff
@@ -511,6 +515,27 @@ export interface InitialAccountCredential {
     contents: InitialCredentialDeploymentValues;
 }
 
+export type BakerPendingChange =
+    | {
+          change: 'ReduceStake';
+          newStake: bigint;
```
orhoj, line 521: You have to make sure this is correctly parsed as a `bigint` when parsing the JSON.

src/types.ts
```diff
@@ -423,9 +424,12 @@ export interface ConsensusStatus {
     finalizationPeriodEMSD?: number;
 
     genesisTime: Date;
+    currentEraGenesisTime: Date;
     blockLastReceivedTime?: Date;
     blockLastArrivedTime?: Date;
     lastFinalizedTime?: Date;
+
+    protocolVersion: number;
```
orhoj, line 432: There is also the field `genesisIndex` that we should add.

src/types.ts
```diff
@@ -393,6 +393,7 @@ export interface BlockInfo {
 export interface ConsensusStatus {
```
orhoj, line 393: Please add `genesisIndex` as well as that is also available.

orhoj:
You should bump the version of the SDK so that we can make a release with the changes afterwards.

Issue #20, opened 2021-11-12
Add support for generating account aliases from an address


abizjak:
**Task description**

As the title states, we need to add a method to generate aliases given an address and a counter.

We should also add a function for checking whether addresses point to the same account.

orhoj:
This was published to npm with version `0.5.1`.

PR #21, opened 2021-11-17
Add isAlias and getAlias functions

shjortConcordium:
## Purpose

Implement #20.

## Changes

Added `isAlias` and `getAlias` functions, and tests for them.
~~`getAlias` is made so it can treat the canonical address as an alias, by starting the counter from 0,
and otherwise start it from 1.~~

## Checklist

      hard-to-understand areas.


abizjak (approved):

src/alias.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+export function getAlias(
+    address: AccountAddress,
+    counter: number
+): AccountAddress {
+    if (counter < 0 || counter > maxCount) {
+        throw new Error('counter exceed maximum number of aliases');
+    }
+    const commonBytes = address.decodedAddress.slice(0, commonBytesLength);
+    const aliasBytes = Buffer.alloc(aliasBytesLength);
```
orhoj, line 46: Please use `buffer/` instead of the node Buffer.

test/alias.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    const address = new AccountAddress(
+        '4hXCdgNTxgM7LNm8nFJEfjDhEcyjjqQnPSRyBS9QgmHKQVxKRf'
+    );
+    const alias = getAlias(address, 1);
+    const otherAlias = getAlias(address, 100);
+    expect(otherAlias.address).not.toBe(alias.address);
+    expect(isAlias(otherAlias, alias)).toBeTruthy();
+});
+
+test('getAlias: last 3 bytes of alias matches counter', () => {
```
orhoj, line 96: I think this test with the minimum and maximum counter values would be good to have too.

orhoj (changes requested):

test/alias.test.ts
shjortConcordium: Added them to that test

orhoj:
Please update the CHANGELOG with an entry for the added functionality.

orhoj (changes requested):
Please update the CHANGELOG with the changes.

orhoj (approved):

PR #27, opened 2022-01-18
Add scheduled and encrypted transfers

shjortConcordium:
## Changes

Added support for scheduled and encrypted transfers (and their memo versions).
Added function to build encrypted transfers.

## Checklist

      hard-to-understand areas.


rust-src/aux_functions.rs
```diff
@@ -7,6 +7,12 @@ use std::collections::BTreeMap;
 type ExampleCurve = G1;
 use serde::{Deserialize as SerdeDeserialize, Serialize as SerdeSerialize};
 use sha2::{Digest, Sha256};
+// Used for shielded transfer:
```
orhoj, line 10: We do not usually comment what the imports are used for like this.

PR #28, opened 2022-02-02
Add deserialization of contract state

shjortConcordium:
## Purpose

Add function to deserialize/convert a contract's state into JSON.

## Changes

Added the function (deserializeContractState) and a test for it.
Added concordium-contracts-common as dependency.
Bumped version to 0.6.0.

## Checklist

      hard-to-understand areas.


rust-src/aux_functions.rs
```diff
@@ -176,3 +178,19 @@ pub fn get_credential_deployment_info_aux(
     let cdi_json = json!(cdi);
     Ok(cdi_json.to_string())
 }
+
+pub fn deserialize_state_aux(
```
Bargsteen, line 182: Please add some doc comments.

rust-src/external_functions.rs
```diff
@@ -33,3 +33,16 @@ pub fn get_credential_deployment_info_ext(
         Err(e) => format!("unable to get credential due to: {}", e),
     }
 }
+
+#[wasm_bindgen(js_name = deserializeState)]
+pub fn deserialize_state(
```
Bargsteen, line 38: Doc comments 😊

src/deserialization.ts
```diff
@@ -0,0 +1,26 @@
...
+        contractName,
+        state.toString('hex'),
+        schema.toString('hex')
+    );
+    try {
+        return JSON.parse(serializedState);
+    } catch (e) {
+        throw new Error(
+            'unable to deserialize state, due to: ' + serializedState
+        ); // In this case the state is the error message from the rust module
```
Bargsteen, line 24:
```suggestion
        ); // In this case the serializedState is the error message from the rust module
```

src/index.ts
```diff
@@ -33,3 +33,4 @@ export {
     buildSignedCredentialForExistingAccount,
 } from './credentialDeploymentTransactions';
 export { isAlias, getAlias } from './alias';
+export { deserializeContractState } from './deserialization';
```
Bargsteen, line 36: The file could also be called something with schema, to clarify what is being deserialized

test/deserialization.test.ts
```diff
@@ -0,0 +1,40 @@
...
+        blockHash,
+        contractAddress
+    );
+    if (!instanceInfo) {
+        throw new Error(
+            'The instance info should exist for the provided block hash.'
+        );
+    }
+    const schema = Buffer.from(
+        fs.readFileSync('./test/resources/two-step-transfer-schema.bin')
```
Bargsteen, line 23: Could we add a test/example of how to use the embedded schema?

Bargsteen (approved):
Looks good overall. Minor comments ;)

test/deserialization.test.ts
shjortConcordium: It seems we can't, because the embedded schema comes from getModuleSource, which we haven't implemented yet (I just realise now). :sweat_smile:

abizjak: That is part of blocktech PR right?

src/index.ts
shjortConcordium: When it is used externally it is all one module, so it's mostly for internal structure, and I think it's fine like this until we get too much de-serialization, and we would need to split it up.

PR #39, opened 2022-04-13
avoid panicking in DeserializeState

shjortConcordium:
## Purpose

When wasm modules panics, the error message is not provided in the calling JS code.
So I changed `deserializeState` to not panic, if it is unable to parse the state with the given schema.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/aux_functions.rs
+++ b/rust-src/aux_functions.rs
@@ -193,6 +193,9 @@ pub fn deserialize_state_aux(
     let mut state_cursor = Cursor::new(hex::decode(state_bytes)?);
     let contract_schema = module_schema.contracts.get(contract_name).ok_or(anyhow!("Unable to get contract schema: not included in module schema"))?;
     let state_schema = contract_schema.state.as_ref().ok_or(anyhow!("Unable to get state schema: not included in contract schema"))?;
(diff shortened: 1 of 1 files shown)


Bargsteen (approved):
LGTM

rust-src/aux_functions.rs
```diff
@@ -193,6 +193,9 @@ pub fn deserialize_state_aux(
     let mut state_cursor = Cursor::new(hex::decode(state_bytes)?);
     let contract_schema = module_schema.contracts.get(contract_name).ok_or(anyhow!("Unable to get contract schema: not included in module schema"))?;
     let state_schema = contract_schema.state.as_ref().ok_or(anyhow!("Unable to get state schema: not included in contract schema"))?;
-    Ok(state_schema.to_json(&mut state_cursor).expect("Unable to parse state to json").to_string())
+    match state_schema.to_json(&mut state_cursor) {
+        Ok(schema) => Ok(schema.to_string()),
+        Err(_) => return Err(anyhow!("Unable to parse state to json"))
```
abizjak, line 198: It would be even more useful if you did not ignore the error here and instead provided it back to the user. It might help debugging. Also, return is not necessary. I'm surprised clippy did not balk at you.

So something like
```
        Err(e) => Err(anyhow!("Unable to parse state to json: {}", e))
```

PR #41, opened 2022-04-22
Restructure Project

shjortConcordium:
## Purpose

Restructure project so we can support both nodejs and web environments.

## Changes

- Split the SDK into a common package and a nodejs package.
- Added a "web" package that can be bundled for web. (Currently just exposes everything from the common package)

## Checklist

      hard-to-understand areas.


package.json
```diff
@@ -1,7 +1,18 @@
 {
-    "name": "@concordium/node-sdk",
-    "version": "0.8.0-alpha.0",
-    "description": "Helpers for interacting with the Concordium node",
+    "private": true,
+    "workspaces": [
+        "./packages/common",
+       "./packages/nodejs",
```
shjortConcordium, line 5: :thinking:

packages/web/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+    "extends": "../../tsconfig-base.json",
+    "include": ["src/**/*"],
+    "exclude": ["node_modules", "grpc/**/*"],
```
soerenbf, line 4: Redundant.

packages/web/package.json
```diff
@@ -0,0 +1,43 @@
...
+        "jest": "^27.0.6",
+        "ts-jest": "^27.0.3",
+        "typescript": "^4.3.5",
+        "webpack": "^5.72.0",
+        "webpack-cli": "^4.9.2",
+        "stream-browserify": "^3.0.0"
+    },
+    "scripts": {
+        "build": "tsc",
+        "postbuild": "yarn webpack",
```
soerenbf, line 38: Shouldn't this just be

```suggestion
        "postbuild": "webpack",
```

packages/web/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Wrappers for interacting with the Concordium node, for the web environment.
+
+## Building for a release
+To build the package run
+```
+yarn build
+```
+
+This builds the [common package](../common), and compiles the project in to javascript. Then it uses webpack to create a single, pure-js file, which is put in dist.
```
soerenbf, line 13:
```suggestion
This builds the [common package](../common), and transpiles the project to javascript. Then it uses webpack to create a single, pure-js file, which is put in dist.
```

packages/web/tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+    "rerefences": [
+        { "path": "../common"}
+    ],
+    "compilerOptions": {
+        "outDir": "./lib/src"
```
soerenbf, line 9: Would't it be nicer with simply

```suggestion
        "outDir": "./lib"
```

packages/nodejs/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+    "extends": "../../tsconfig-base.json",
+    "include": ["src/**/*"],
+    "exclude": ["node_modules", "grpc/**/*"],
+    "rerefences": [
+        { "path": "../common"}
+    ],
+    "compilerOptions": {
+        "rootDir": "src",
+        "outDir": "./lib/src"
```
soerenbf, line 10: Similar to web package, would it be nicer with

```suggestion
        "outDir": "./lib"
```

soerenbf (dismissed):
Readme could do with an installation section. I was an idiot and forgot submodules..

Furthermore, consider the following comments

packages/common/package.json
```diff
@@ -0,0 +1,55 @@
+{
+    "name": "@concordium/common",
```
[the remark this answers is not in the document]
shjortConcordium: Yeah, I'm generally not sure about the name. But I would be fine changing it as suggested

packages/nodejs/package.json
```diff
@@ -0,0 +1,56 @@
+{
+    "name": "@concordium/node-sdk",
```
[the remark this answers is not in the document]
shjortConcordium: There are people using this already, so I would prefer not to rename it :shrug:

packages/nodejs/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+    "extends": "../../tsconfig-base.json",
+    "include": ["src/**/*"],
+    "exclude": ["node_modules", "grpc/**/*"],
+    "rerefences": [
+        { "path": "../common"}
```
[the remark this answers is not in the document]
shjortConcordium: I assume that it uses the outDir. (At least that seems to be the behaviour)
 
 > Where will the common stuff be accessible for a consumer of the published module?
 That is something to be decided on, I guess. Currently I just reexported everything from common, so that we could update without a user noticing.

packages/nodejs/tsconfig.json
shjortConcordium: I guess we should be alert doing this, it might break something :upside_down_face:

packages/nodejs/package.json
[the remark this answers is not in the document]
soerenbf: Alright.. I guess if we have something called `web-sdk` as well, it should be possible to figure out.

packages/web/package.json
```diff
@@ -0,0 +1,43 @@
+{
+    "name": "@concordium/web",
```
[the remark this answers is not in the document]
soerenbf: Yes... do you like `web3`? I personally think it doesn't say anything about the scope of the package, but if anyone has a strong opinion I don't care. I think with the `sdk` postfix, we align with our other releases instead of aligning with the general blockchain community, so that's the stuff we have to weigh.

packages/nodejs/tsconfig.json
soerenbf: I guess it would. But we're still in the beta versions, aren't we?

packages/nodejs/tsconfig.json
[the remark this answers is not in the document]
soerenbf: Yes, I think I was thinking of a case where `common` wouldn't be published as a separate package or something.. Makes sense now at least! :smile:

packages/common/package.json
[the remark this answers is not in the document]
orhoj: We currently have:
- node-sdk

So either we remain consistent with that:

- node-sdk
- web-sdk
- common-sdk

or we find a new way to name the packages.

packages/common/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "version": "0.0.1",
```
orhoj, line 3: I suggest that we start at `1.0.0` as everything in this package has already been published before as part of node-sdk. Or at the very least `0.1.0` :smile:

packages/common/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": "../../tsconfig-base.json",
+    "include": ["src/**/*"],
+    "compilerOptions": {
+        "outDir": "./lib/src"
```
orhoj, line 5: The other packages now use `lib` as the outDir. Are you keeping this to try to avoid breaking stuff?

packages/web/package.json
```diff
@@ -0,0 +1,46 @@
+{
+    "name": "@concordium/web-sdk",
+    "version": "0.0.1",
```
orhoj, line 3:
```suggestion
    "version": "0.1.0",
```

orhoj (changes requested):

.github/workflows/pipeline.yml
```diff
@@ -31,6 +31,7 @@ jobs:
               run: yarn
 
             - name: Build gRPC
+              working-directory: packages/common
```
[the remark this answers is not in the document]
shjortConcordium: No, it is only for the gRPC step. (which is only in, now, nodejs)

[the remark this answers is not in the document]
shjortConcordium: Maybe I could remove this step, because it should be done automatically in preinstall :thinking:

packages/common/package.json
shjortConcordium: I'll set it to 1.0.0 :+1:

packages/common/tsconfig.json
shjortConcordium: I'll change it to lib, like the others.

README.md
```diff
@@ -933,25 +32,13 @@ yarn build
...
-
-## Updating the gRPC files
-If the external dependency concordium-grpc-api has been updated (or this is your first time building the project), 
-then it is required to regenerate the files from the `.proto` file. Do this by running:
-```
-yarn generate
-```
-This will overwrite the existing files in `src/grpc/`. Remember to check that existing functionality still
-works after performing an update.
+Currently publishing is done indiviudally for each package.
```
orhoj, line 35:
```suggestion
Currently publishing is done individually for each package. 
```

orhoj (changes requested):
If I run
```
yarn test
```
then it fails.

packages/nodejs/README.md
```diff
@@ -0,0 +1,512 @@
...
+
+# Build
+
+## Building for a release
+To build the package run
+```
+yarn build
+```
+
+Note that this also builds the [common package](../common).
```
orhoj, line 497: This does not seem to be the case. If I run `yarn build` for the `nodejs` package, then it fails if `common` has not been built.

shjortConcordium: Yeah, I changed the text, to say you need to built it first.

shjortConcordium:
> If I run
> 
> ```
> yarn test
> ```
> 
> then it fails.

They should run now, though they fail because the nodes have been updated.

packages/rust-bindings/README.md
```diff
@@ -0,0 +1,21 @@
+# concordium-rust-bindings
+
+[![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.0-4baaaa.svg)](https://github.com/Concordium/.github/blob/main/.github/CODE_OF_CONDUCT.md)
+
+Bindings for rust function, that is to be used in the sdk's.
```
orhoj, line 5:
```suggestion
Bindings for the Rust functions that are used by the SDK.
```

packages/rust-bindings/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Building for a release
+To build the package run
+```
+yarn build
+```
+
+This compiles the rust into wasm with javascript wrappers.
```
orhoj, line 13:
```suggestion
This compiles the Rust code into wasm with Javascript wrappers.
```

packages/rust-bindings/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Publishing a release
+Before publishing a new release it is essential that it has been built first. So make sure that 
+you have just built the up-to-date code you want to publish. To publish the release run
+```
+yarn publish
+```
+and step through the steps precented to you.
```
orhoj, line 21:
```suggestion
and step through the steps presented to you.
```

packages/rust-bindings/package.json
```diff
@@ -0,0 +1,24 @@
...
+        "node": ">=14.16.0"
+    },
+    "main": "pkg/node/concordium_rust_bindings.js",
+    "module": "pkg/bundler/concordium_rust_bindings.js",
+    "types": "pkg/bundler/concordium_rust_bindings.d.ts",
+    "files": [
+        "/pkg/**/*"
+    ],
+    "devDependencies": {
+        "cross-env": "5.0.5"
```
orhoj, line 16: Are you using this dependency?

packages/web/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Wrappers for interacting with the Concordium node, for the web environment.
+
+## Building for a release
+To build the package run
+```
+yarn build
+```
+
+This transpiles the project to javascript. Then it uses webpack to create a single, pure-js file, which is put in dist.
```
orhoj, line 13:
```suggestion
This transpiles the project to Javascript, wherefter webpack is used to create a single pure Javascript file which is put in the `lib` directory.
```

packages/web/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Note that the dependent packages must already have been built. To easily do this, build from the package root instead.
+
+## Publishing a release
+Before publishing a new release it is essential that it has been built first. So make sure that 
+you have just built the up-to-date code you want to publish. To publish the release run
+```
+yarn publish
+```
+and step through the steps precented to you.
```
orhoj, line 23:
```suggestion
and step through the steps presented to you.
```

packages/web/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "author": {
+        "name": "Concordium Software",
+        "email": "support@concordium.software",
+        "url": "https://concordium.com"
+    },
+    "devDependencies": {
+        "@typescript-eslint/eslint-plugin": "^4.28.1",
+        "@typescript-eslint/parser": "^4.28.1",
+        "babel-jest": "^27.0.6",
+        "cross-env": "5.0.5",
```
orhoj, line 24: It looks like this is unused.

packages/web/example/Alias.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8">
+        <title></title>
+        <script src="../dist/concordium.min.js"></script>
```
orhoj, line 6:
```suggestion
        <script src="../lib/concordium.min.js"></script>
```

packages/web/example/Alias.html
```diff
@@ -0,0 +1,30 @@
...
+        <script  type="module" >
```
orhoj, line 7:
```suggestion
        <script type="module" >
```

packages/web/webpack.config.js
```diff
@@ -37,26 +36,26 @@ module.exports = {
...
         asyncWebAssembly: true,
     },
     output: {
-        filename: "[name].min.js",
-        path: path.resolve(__dirname, "lib"),
-        library: "concordium",
-        libraryTarget: "umd",
+        filename: '[name].min.js',
+        path: path.resolve(__dirname, 'lib'),
+        library: 'concordiumSDK',
```
orhoj, line 57: I wonder if there is a naming convention for library names, e.g. should it be `concordiumsdk` instead?

orhoj (approved):

packages/web/webpack.config.js
shjortConcordium: Looking into it a bit, I can't say I can find a standard. I did find that you can do:
library: ['concordium', 'sdk']
and it should be available on `window.concordium.sdk`.
(But that would only work if we had the wallet api on `window.concordium.wallet`, and they would not overwrite each other.

PR #53, opened 2022-05-23
JSON Rpc client

shjortConcordium:
## Purpose

- Create a client for json-rpc.

## Changes

- Create json-rpc client. (With 4 calls)
- Create http "provider" for json-rpc client.
- Added helpers for signing account transactions.

## Checklist

      hard-to-understand areas.


packages/common/src/providers/provider.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export interface JsonRpcResponse {
+    error?: {
+        code: number;
+        message: string;
+        data: any;
+    };
+    result: any;
```
orhoj, line 9: The specification states that you only ever have `error` or `result`, but never both at the same time. We should improve the typing here.

packages/common/src/providers/provider.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+export interface JsonRpcResponse {
+    error?: {
+        code: number;
+        message: string;
+        data: any;
```
orhoj, line 7: `data` is optional in an `error`.

packages/common/src/providers/httpProvider.ts
```diff
@@ -0,0 +1,39 @@
+import Provider, { JsonRpcResponse } from './provider';
+import fetch from 'cross-fetch';
+
+export class HttpProvider implements Provider {
+    request: (
+        method: string,
+        params?: Record<string, unknown>
+    ) => Promise<JsonRpcResponse>;
+    nextId = 0;
```
orhoj, line 9: I believe using a `UUID` instead of an incrementing number as the `id` in the requests is better. If you reset the client using this class, then it will start using the same `id` values making it harder to search through requests based on a specific `id`.

packages/common/src/providers/httpProvider.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    constructor(url: string) {
+        this.request = async function (
+            method: string,
+            params?: Record<string, unknown>
+        ) {
+            const request = {
```
orhoj, line 16: We have a nice type for the response, so we should also have one for the request.

packages/common/src/providers/httpProvider.ts
```diff
@@ -0,0 +1,39 @@
...
+                method: 'POST',
+                body: JSON.stringify(request),
+                headers: {
+                    'Content-Type': 'application/json',
+                },
+            };
+
+            const res = await fetch(url, options);
+            if (res.status >= 400) {
+                throw new Error('Bad response from server');
```
orhoj, line 33: The error should extract the error information received from the server. This error will not help a user in debugging what went wrong (i.e. they provided incorrect parameters or whatever).

packages/web/webpack.config.js
```diff
@@ -10,6 +10,9 @@ module.exports = {
         new webpack.SourceMapDevToolPlugin({
             filename: '[file].map',
         }),
+        new webpack.ProvidePlugin({
+            process: 'process/browser',
```
orhoj, line 14: I don't see what this is being used for.

packages/common/src/JsonRpcClient.ts
```diff
@@ -0,0 +1,195 @@
...
+    ): Promise<NextAccountNonce | undefined> {
+        const response = await this.provider.request('getNextAccountNonce', {
+            address: accountAddress.address,
+        });
+
+        const bigIntPropertyKeys: (keyof NextAccountNonce)[] = ['nonce'];
+
+        return handleResponse(response, (r) =>
+            transformJsonResponse(
+                JSON.stringify(r),
```
orhoj, line 67: This looks problematic. What happens if there is a `bigint` in `r` when doing this?

packages/common/src/signHelpers.ts
```diff
@@ -0,0 +1,82 @@
...
+                },
+            };
+        },
+    };
+}
+
+export type CredentialKeyStructure = Record<number, string>;
+export type AccountKeyStructure = Record<number, CredentialKeyStructure>;
+
+export function buildAdvancedAccountSigner(
```
orhoj, line 37: Remove this as we are not going to use it.

packages/common/src/signHelpers.ts
```diff
@@ -0,0 +1,82 @@
...
+import { countSignatures } from './util';
+import * as ed from 'noble-ed25519';
+import { Buffer } from 'buffer/';
+
+export interface AccountSigner {
+    sign(digest: Buffer): Promise<AccountTransactionSignature>;
+    getSignatureCount(): bigint;
+}
+
+export function buildBasicAccountSigner(privateKey: string): AccountSigner {
```
orhoj, line 16: Document exported functions, e.g. it is important to state the format of `privateKey`.

packages/web/webpack.config.js
shjortConcordium: We have some dependencies that use process.nextTick. This polyfills that.
(I don't remember which though)

packages/common/src/JsonRpcClient.ts
shjortConcordium: Yeah, it was lossy, I've reworked the handling, to avoid a lossy parse before this.

packages/common/src/providers/httpProvider.ts
shjortConcordium: I tried creating one.

orhoj: But you didn't use it here? I would expect `request` to have an explicit type :smile:

packages/web/webpack.config.js
shjortConcordium: It is _stream_readable.js, which is the polyfill for the nodejs stream readable.

orhoj (approved):

PR #56, opened 2022-06-03
Fix intToString

### LINKED ISSUE #54: Fix intToString
**Bug Description**

`util.js`'s `intToString` doesn't work (as in it doesn't do anything)
This function is used to parse uint64 types from the node.
The impact of this is that we don't convert number fields into string before letting JS parse it, which means that we have incorrect/lossy parsing of numbers above MAX_SAFE_INTEGER.

shjortConcordium:
## Purpose

Fix #54

## Changes

intToString now returns the modified object, instead of the input.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -13,9 +13,9 @@ import { AccountTransactionSignature } from './types';
  * @returns the same JSON string where the numbers at the supplied keys are quoted
  */
 function intToString(jsonStruct: string, keys: string[]): string {
-    const result = jsonStruct;
+    let result = jsonStruct;
     for (const key of keys) {
-        result.replace(
+        result = result.replace(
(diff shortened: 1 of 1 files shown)


src/util.ts
```diff
@@ -13,9 +13,9 @@ import { AccountTransactionSignature } from './types';
  * @returns the same JSON string where the numbers at the supplied keys are quoted
  */
 function intToString(jsonStruct: string, keys: string[]): string {
-    const result = jsonStruct;
+    let result = jsonStruct;
     for (const key of keys) {
-        result.replace(
+        result = result.replace(
```
orhoj, line 18: Add a unit test or two that demonstrates that this now works as expected.

orhoj (changes requested):
Make the CI pipeline run towards release branches.

orhoj (approved):

PR #62, opened 2022-06-14
Add function to deserialize schema's for v1 contracts

### LINKED ISSUE #61: Add support for schema version 2 (for v1 contracts)
**Task description**
Add support for schema version 2, which matches smart contracts version 1.
Please note that version 1 of the schema should still be supported.

In schema version 2, there is no longer a schema for the state and each function can have optional schemas for a parameter and return value.
Please see https://github.com/Concordium/concordium-contracts-common/blob/020bac62e18eac03af50b36e76b5a4b81eadad15/src/schema.rs#L60 for more details.

shjortConcordium:
## Purpose

Fix #61 

## Changed

- Support deserializing version 2 schemas.
- Support serializing parameters for V1 contracts, by adding a another parameter to denote module version.

## Checklist

      hard-to-understand areas.


README.md
```diff
@@ -872,6 +882,16 @@ const inputParams = serializeUpdateContractParameters(
         modulefileBuffer
 );
 ```
+Note that if using a V1 contract, the 5th argument to `serializeUpdateContractParameters` should be 1:
```
Bargsteen, line 885: We are using semantic versioning for the schemas (although we only used the major version so far), so while this is fine atm. it won't be in the future.
We decided that schema v3 will include the version directly in the schema (as opposed to only in the wasm module). That will alleviate some of the problems you are facing here.
Although v1 and v2 schemas should still be supported by our tools in the future.

src/deserializeSchema.ts
```diff
@@ -166,6 +207,27 @@ export type PairType = {
...
+            return {
+                returnValue: deserialType(source),
+            };
+        case 2:
+            return {
+                parameter: deserialType(source),
+                returnValue: deserialType(source),
+            };
+        default:
+            throw new Error('Incorrect index for function');
```
Bargsteen, line 227:
```suggestion
            throw new Error('Incorrect tag for function');
```
We usually call them tags on the rust/haskell side. Is index used in JS?

src/deserializeSchema.ts
```diff
@@ -466,14 +528,36 @@ export function deserialUint32(source: Readable): number {
...
+): Module;
+export function deserialModuleFromBuffer(
+    buffer: Buffer,
+    moduleVersion: 1
+): ModuleV1;
+export function deserialModuleFromBuffer(
+    buffer: Buffer,
+    moduleVersion: number
+): Module | ModuleV1;
+export function deserialModuleFromBuffer(
```
Bargsteen, line 547: I understand that you want to leave `0` as the default to avoid a major version change, but concordium-client and cargo-concordium use smart contract version 1 by default, so this might be confusing for the users.
Consider making this a breaking change.

shjortConcordium: Yeah, we'll do this as a major.
Though I have made the decision to not use a default. Then when we add support for higher versions, then we can have that as the default.
Also I assume this is not really used directly, but instead through the `serializeInitContractParameters` and `serializeUpdateContractParameters` functions.

src/deserializeSchema.ts
shjortConcordium: I renamed it to tag. Honestly didn't think about it and copied the var name from the rust implementation, where it was idx (I assume that's short for index :upside_down_face: )

packages/common/src/deserializeSchema.ts
```diff
@@ -164,6 +228,42 @@ export type PairType = {
     ofRight: Type;
 };
 
+/**
+ * The first byte of the function denotes what whether the parameters/return value or both are typed.
```
Bargsteen, line 232:
```suggestion
 * The first byte of the function denotes whether the parameters, return value, or both are typed.
```

packages/common/src/serialization.ts
```diff
@@ -432,62 +436,61 @@ export function serializeCredentialDeploymentTransactionForSubmission(
 }
 
 /**
- *
  * @param contractName name of the contract that the init contract transaction will initialize
- * @param userInput  user input
- * @param modulefileBuffer buffer of embedded schema file
+ * @param parameters  the parameters to be serialized, should correspond to the JSON representation.
```
Bargsteen, line 440:
```suggestion
 * @param parameters the parameters to be serialized. Should correspond to the JSON representation.
```

packages/common/src/serialization.ts
```diff
@@ -432,62 +436,61 @@ export function serializeCredentialDeploymentTransactionForSubmission(
...
+    return serializeParameters(initType, parameters);
 }
 
 /**
- *
  * @param contractName name of the contract that the update contract transaction will update
  * @param receiveFunctionName name of function that the update contract transaction will invoke
- * @param userInput user input
- * @param modulefileBuffer buffer of embedded schema file
+ * @param parameters  the parameters to be serialized, should correspond to the JSON representation
```
Bargsteen, line 468:
```suggestion
 * @param parameters the parameters to be serialized. Should correspond to the JSON representation.
```

Bargsteen (approved):
LGTM 👍

PR #69, opened 2022-06-27
V2 fixes

shjortConcordium:
## Purpose

Fix errors of current SDK versions.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/nodejs/package.json
+++ b/packages/nodejs/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@concordium/node-sdk",
-    "version": "2.0.0",
+    "version": "2.0.1",
     "description": "Helpers for interacting with the Concordium node",
     "repository": {
         "type": "git",
@@ -18,6 +18,7 @@
     "main": "lib/index.js",
     "types": "lib/index.d.ts",
     "files": [
+        "/grpc/*",
         "/lib/**/*"
     ],
     "devDependencies": {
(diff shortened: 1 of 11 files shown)


packages/nodejs/src/util.ts
```diff
@@ -37,3 +38,12 @@ export function unwrapJsonResponse<T>(
 
     return JSON.parse(jsonString, reviver);
 }
+
+/**
+ * Loads the module as a buffer, given the given filePath.
+ * @param filepath the location of the module
+ * @returns the module as a buffer
+ */
+export function getModuleBuffer(filePath: string): Buffer {
```
orhoj, line 47: I am not sure that I understand the need for this method.

packages/web/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Changelog
 
-## 0.2.0
+## 0.2.1 2022-6-27
+
+- Fixed errors in common dependency.
```
orhoj, line 5: Please follow the format:
```
### Fixed
- Errors in common dependency.
```
Also, it would be better to be more exact about what error that was.

packages/nodejs/src/util.ts
orhoj: We will keep it there, but in general I would advise users to code their own file system handling.

orhoj:
We are missing
```
google-protobuf
```
from the dependencies as well.

packages/nodejs/package.json
```diff
@@ -51,10 +52,10 @@
         "lint": "eslint . --cache --ext .ts,.tsx --max-warnings 0",
         "lint-fix": "yarn --silent lint --fix; exit 0",
         "test": "jest",
-        "build": "([ ! -e \"grpc\" ] && yarn generate) || true && tsc && cp -r ./grpc/ ./lib/ "
+        "build": "([ ! -e \"grpc\" ] && yarn generate) || true && tsc"
     },
     "dependencies": {
```
orhoj, line 57: We are missing `google-protobuf` from the dependencies.

packages/web/CHANGELOG.md
shjortConcordium: I have clarified.

orhoj (approved):

PR #79, opened 2022-08-03
Add function to create new identities

shjortConcordium:
## Purpose

Add function to create V1 Identity requests.

## Changes

- Added the `createIdentityRequest` function.
- Added a test

## Checklist

      hard-to-understand areas.


packages/common/src/identity.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Creates an V1 identityRequest.
+ */
+export function createIdentityRequest(
+    input: IdentityRequestInput
+): Versioned<IdObjectRequest> {
+    const rawRequest = wasm.createIdRequestV1(JSON.stringify(input));
+    try {
+        return JSON.parse(rawRequest).idObjectRequest;
+    } catch (e) {
+        throw new Error(rawRequest);
```
orhoj, line 42: Will `e` not hold any useful information if something fails?

packages/common/src/identity.ts
```diff
@@ -0,0 +1,44 @@
...
+    identityIndex: number;
+    arThreshold: number;
+};
+
+export type IpArData = {
+    encPrfKeyShare: string;
+    proofComEncEq: string;
+};
+
+export interface IdObjectRequest {
```
orhoj, line 19: Should this type be called `V1` since it is generated from `createIdRequestV1`?

packages/common/test/id.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        arsInfos,
+        seed,
+        net: 'Testnet',
+        identityIndex: 0,
+        arThreshold: 1,
+    };
+    const output = createIdentityRequest(input).value;
+    const output2 = createIdentityRequest(input).value;
+    expect(typeof output.proofsOfKnowledge).toBe('string');
+    expect(output.idCredPub).toEqual(output2.value.idCredPub);
```
orhoj, line 29: Hmm?

packages/nodejs/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- `createIdentityRequest`, to create identity requests.
```
orhoj, line 7: I think we should update the CHANGELOG as described here: https://github.com/Concordium/concordium-node-sdk-js/blob/7b8d693cc7e0b8daa765efc0dd96d81c9ba1c73d/README.md instead. Then we can avoid duplicating the CHANGELOG on every update to common.

packages/web/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 
+- `createIdentityRequest`, to create identity requests.
```
orhoj, line 7: Same as for the nodejs CHANGELOG.

packages/rust-bindings/src/aux_functions.rs
```diff
@@ -24,6 +33,88 @@ pub struct CredId {
...
+
+    let id_cred_sec: PedersenValue<ArCurve> =
+        PedersenValue::new(wallet.get_id_cred_sec(input.identity_index)?);
+    let id_cred: IdCredentials<ArCurve> = IdCredentials { id_cred_sec };
+
+    let sig_retrievel_randomness: ps_sig::SigRetrievalRandomness<Bls12> =
+        wallet.get_blinding_randomness(input.identity_index)?;
+
+    let num_of_ars = input.ars_infos.len();
+    let threshold = match input.ar_threshold {
```
orhoj, line 72: The `threshold` should be required in the input instead of being optional.

packages/rust-bindings/src/aux_functions.rs
```diff
@@ -24,6 +33,88 @@ pub struct CredId {
...
+    };
+
+    let chi = CredentialHolderInfo::<ArCurve> { id_cred };
+
+    let aci = AccCredentialInfo {
+        cred_holder_info: chi,
+        prf_key,
+    };
+
+    // Choice of anonymity revokers, all of them in this implementation.
```
orhoj, line 99: This comment does not seem to fit in here.

packages/rust-bindings/src/aux_functions.rs
```diff
@@ -97,6 +188,7 @@ pub fn generate_unsigned_credential_aux(input: &str) -> Result<String> {
         policy,
         cred_key_info,
         address.as_ref(),
+        &SystemAttributeRandomness {},
```
orhoj, line 191: Did the interface of `create_unsigned_credential` change so that this change is necessary?

shjortConcordium: Yes, it was changed to allow the randomness to be chosen. This just means that the function will use whatever the system offers. (Which is what it did before)

packages/common/src/identity.ts
shjortConcordium: The rust code will either return a proper json, or an error message. So `e` should never old any useful information.

packages/common/test/id.test.ts
shjortConcordium: I agree. I changed it to test the structure, and the values for the fields that as deterministic.

packages/rust-bindings/src/aux_functions.rs
shjortConcordium: Yeah, I removed it. It's just the notion that we are putting all ars' infos into the context. But if we were to change this, it would be outside the rust code anyway.

packages/web/CHANGELOG.md
```diff
@@ -2,9 +2,9 @@
 
 ## 0.4.0 2022-8-8
 
-### Added
+### Changed
 
-- `deserializeTransaction` function to deserialize transaction created by `serializeAccountTransactionForSubmission` and `serializeCredentialDeploymentTransactionForSubmission`. (Currently SimpleTransfer, SimpleTransferWithMemo and RegisterData are the only supported account transactions kinds)
+- Bumped @concordium/common-sdk to 2.4.0.
```
orhoj, line 7: This is duplicate.

shjortConcordium: Yes.

orhoj (approved):

PR #80, opened 2022-08-03
Add function to create credentials with seed.

shjortConcordium:
## Purpose

Add function to create credentials with V1 algorithm + allow sending credentials with JSON-RPC clientt

## Changes

- Added the `createCredentialV1` function.
- Added a test.

## Checklist

      hard-to-understand areas.


packages/common/src/JsonRpcClient.ts
```diff
@@ -86,6 +90,16 @@ export class JsonRpcClient {
         return res.result;
     }
 
+    /**
+     * @param serializedTransaction the transaction serialized as a base64-encoded string.
+     */
+    async sendRawTransaction(serializedTransaction: string): Promise<boolean> {
```
orhoj, line 96: Make this private.

packages/common/src/JsonRpcClient.ts
```diff
@@ -86,6 +90,16 @@ export class JsonRpcClient {
...
+        const res = await this.provider.request('sendAccountTransaction', {
```
orhoj, line 97: We should update the name from `sendAccountTransaction` to `sendTransaction` if the JSON-RPC endpoint supports all transaction types anyway.

packages/common/src/credentialDeploymentTransactions.ts
```diff
@@ -177,3 +182,35 @@ export function getAccountAddress(credId: string): AccountAddress {
...
     return accountAddress;
 }
+
+export type CredentialInputV1 = {
+    ipInfo: IpInfo;
+    globalContext: CryptographicParameters;
+    arsInfos: Record<string, ArInfo>;
+    idObject: IdentityObjectV1;
+    revealedAttributes: AttributeKey[];
+    seed: string;
```
orhoj, line 192: We could improve the naming here so that it is clear what format the seed has, e.g. `seedAsHex` or `seedAsMnemonic`.

packages/common/src/credentialDeploymentTransactions.ts
```diff
@@ -177,3 +182,35 @@ export function getAccountAddress(credId: string): AccountAddress {
...
+    net: 'Testnet' | 'Mainnet';
```
orhoj, line 193: I believe this definition was also added for the identity type, so now we should have a shared type.

orhoj:
Missing updates to CHANGELOG(s)

packages/common/src/JsonRpcClient.ts
shjortConcordium: https://github.com/Concordium/concordium-json-rpc/pull/16

orhoj (approved):

PR #84, opened 2022-08-11
Add createIdentityRecoveryRequest

shjortConcordium:
## Purpose

Add `createIdentityRecoveryRequest` for identity recovery.

## Changes

Added `Add createIdentityRecoveryRequest` and a test.

## Checklist

      hard-to-understand areas.


packages/common/src/identity.ts
```diff
@@ -31,3 +32,29 @@ export function createIdentityRequest(
...
     }
 }
+
+export type IdentityRecoveryRequestInput = {
+    ipInfo: IpInfo;
+    globalContext: CryptographicParameters;
+    seedAsHex: string;
+    net: Network;
+    identityIndex: number;
+    // TODO: change to bigint as the type is u64
```
orhoj, line 42: Yes, please.

packages/common/test/id.test.ts
```diff
@@ -41,3 +46,32 @@ test('idrequest', () => {
...
+    const ipInfo = JSON.parse(
+        fs.readFileSync('./test/resources/ip_info.json').toString()
+    ).value;
+    const globalContext = JSON.parse(
+        fs.readFileSync('./test/resources/global.json').toString()
+    ).value;
+    const seedAsHex =
+        'efa5e27326f8fa0902e647b52449bf335b7b605adc387015ec903f41d95080eb71361cbc7fb78721dcd4f3926a337340aa1406df83332c44c1cdcfe100603860';
+
+    const timestamp = Math.floor(Date.now() / 1000);
```
orhoj, line 60: Just a note that in general it can be dangerous to use dynamic dates in unit tests. The unit tests will run on different input on each run, so it is not consistently testing the same thing. It is generally better to hardcode it instead (if possible).

packages/common/src/identity.ts
shjortConcordium: I removed it instead :upside_down_face:

packages/common/test/id.test.ts
shjortConcordium: Hardcoded it.

orhoj (approved):

PR #85, opened 2022-08-25
Remove contract common add new schema version

shjortConcordium:
## Purpose

Remove direct usage of contracts-common.
Support de-serialization of new schemas with error types.

## Checklist

      hard-to-understand areas.


Bargsteen (approved):
Looks good! 👍 
Perhaps we should use the schemas to parse the returned bytes from ContractInvoke. As far as I can see, the schemas are not used for that yet. But that can wait for another time.

shjortConcordium:
> Looks good! +1 Perhaps we should use the schemas to parse the returned bytes from ContractInvoke. As far as I can see, the schemas are not used for that yet. But that can wait for another time.

I think we need to add a general function to parse using the schemas, but I'll add it as an issue.

PR #88, opened 2022-09-23
Use proper format for signMessage

shjortConcordium:
## Purpose

Replace temporary prepend in signMessage with a proper format.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/test/signHelpers.test.ts
+++ b/packages/common/test/signHelpers.test.ts
@@ -5,6 +5,25 @@ import {
 } from '../src/signHelpers';
 import { AccountAddress, AccountInfo } from '../src';
 
+const TEST_CREDENTIALS = {
+    0: {
+        value: {
+            contents: {
+                credentialPublicKeys: {
+                    keys: {
+                        0: {
+                            verifyKey:
+                                'fef5414fc757cd4694bf0c7ea436f015cb7f87a80d08e1d1085b9cc91f13f376',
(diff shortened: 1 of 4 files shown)


packages/common/test/signHelpers.test.ts
```diff
@@ -24,10 +28,12 @@ test('test verifyMessageSignature', async () => {
         message,
         {
             0: {
-                0: '5420cc5a6c956e8a9b47a834d8aaee1176349435f8d172f0255db28a96040eed553d18cd98e2c873bf37a0c1794d382397f68a230485efab1ef16f678008080a',
+                0: '445197d79ca90d8cc8440328dac9f307932ade0c03cc7aa575b59b746e26e5f1bca13ade5ff7a56e918ba5a32450fdf52b034cd2580929b21213263e81f7f809',
             },
         },
         {
+            accountAddress:
+                '3eP94feEdmhYiPC1333F9VoV31KGMswonuHk5tqmZrzf761zK5',
```
orhoj, line 36: It would be nice with a test where the `accountAddress` here is incorrect, to verify that it fails to verify the signature in that case.

shjortConcordium: Yes. Added one with the wrong accountAddress, and one where the signature is incorrect.

orhoj (approved):

PR #92, opened 2022-10-11
WIP

bisgardo:
## Purpose

When using Wallet Connect, dApps need to be able to send an encoded transaction without signature to the wallet such that it can sign it. The ideal solution would be to send it as plain stringified JSON, but sending it encoded fits better with existing code.

## Changes

Extract a function from existing serialization code to only serialize the header, type, and payload parts of the transaction.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/src/serialization.ts
+++ b/packages/common/src/serialization.ts
@@ -97,39 +97,26 @@ export function serializeAccountTransactionSignature(
 }
 
 /**
- * Serializes a transaction and its signatures. This serialization when sha256 hashed
- * is considered as the transaction hash, and is used to look up the status of a
- * submitted transaction.
+ * Serializes an account transaction into a buffer.
  * @param accountTransaction the transaction to serialize
- * @param signatures signatures on the signed digest of the transaction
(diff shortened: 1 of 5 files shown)


bisgardo:
This is a preliminary review, the code has not been tested.

I first renamed some functions to make them consistently have "signed" in the name if they used a signature (and vice versa). This was reverted in 8cbff38bf25563cf454c67a79fbef97d0f227ce7.

packages/common/src/serialization.ts
```diff
@@ -97,39 +97,26 @@ export function serializeAccountTransactionSignature(
 }
 
 /**
- * Serializes a transaction and its signatures. This serialization when sha256 hashed
- * is considered as the transaction hash, and is used to look up the status of a
- * submitted transaction.
+ * Serializes an account transaction into a buffer.
```
shjortConcordium, line 100:
```suggestion
 * Serializes an unsigned account transaction into a buffer.
```

shjortConcordium (commented):
Seems good so far.

rasmus-kirk:
This seems abandoned, I'll close it

PR #93, opened 2022-10-17
Added deserialization of return values

rasmus-kirk:
## Purpose
Implementing a deserialization method for return values as discussed in this issue

## Changes
Implementation of a auxiliary rust function, a ts function and a ts test.

## Checklist

- [ ] The code compiles without warnings. (They are unrelated)
      hard-to-understand areas.

### DIFF
--- a/packages/common/src/deserialization.ts
+++ b/packages/common/src/deserialization.ts
@@ -212,7 +212,7 @@ export function deserializeTransaction(
  * @param schemaVersion The schema version as a number. This parameter is optional,
  * if you provide a serialized versioned schema this argument won't be needed.
  */
-export function deserializeReturnValue(
+export function deserializeReceiveReturnValue(
     returnValueBytes: Buffer,
     moduleSchema: Buffer,
     contractName: string,
@@ -220,7 +220,7 @@ export function deserializeReturnValue(
     schemaVersion?: number
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
 ): any {
-    const deserializedReturnValue = wasm.deserializeReturnValue(
+    const deserializedReturnValue = wasm.deserializeReceiveReturnValue(
         returnValueBytes.toString('hex'),
         moduleSchema.toString('hex'),
         contractName,
--- a/packages/common/src/deserialization.ts
+++ b/packages/common/src/deserialization.ts
@@ -204,7 +204,7 @@ export function deserializeTransaction(
 }
 
 /**
- * Deserializes a return value from a sequence of bytes into a json object.
+ * Deserializes a receive functions's return value from a sequence of bytes into a json object.
  * @param returnValueBytes A buffer containing the return value as raw bytes.
  * @param moduleSchema The raw module schema as a buffer.
  * @param contractName The name of the contract where the receive function is located.
@@ -233,6 +233,6 @@ export function deserializeReceiveReturnValue(
         throw new Error(
             'unable to deserialize the return value, due to: ' +
                 deserializedReturnValue
-        ); // In this case serializedState is the error message from the rust module
+        ); // In this case deserializedReturnValue is the error message from the rust module
     }
 }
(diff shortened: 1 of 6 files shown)


packages/common/src/deserialization.ts
```diff
@@ -202,3 +202,35 @@ export function deserializeTransaction(
...
+ * @param schemaVersion The schema version as a number. This parameter is optional,
+ * if you provide a serialized versioned schema this argument won't be needed.
+ */
+export function deserializeReturnValue(
+    returnValueBytes: Buffer,
+    moduleSchema: Buffer,
+    contractName: string,
+    functionName: string,
+    schemaVersion?: number
+): any {
```
shjortConcordium, line 221: I think that the lint failing is just because of this.
You should add a:
```suggestion
// eslint-disable-next-line @typescript-eslint/no-explicit-any
): any {
```
To appease it. (We normally don't want any types around, but the type here is pretty much any :upside_down_face: )

packages/rust-bindings/src/aux_functions.rs
```diff
@@ -600,3 +607,80 @@ pub fn deserialize_state_aux(
         Err(e) => Err(anyhow!("Unable to parse state to json: {:?}", e)),
     }
 }
+
+fn get_return_value_schema(
```
shjortConcordium, line 611: Will this be moved to contracts-common?

packages/common/test/deserialization.test.ts
```diff
@@ -122,3 +123,17 @@ test('Expired transactions can be deserialized', () => {
         new TransactionExpiry(new Date(2000, 1), true)
     );
 });
+
+test('Return value can be deserialized', () => {
```
shjortConcordium, line 127: It would be nice to have another test with another contract/version.

packages/common/src/deserialization.ts
```diff
@@ -202,3 +202,37 @@ export function deserializeTransaction(
...
+/**
+ * Deserializes a return value from a sequence of bytes into a json object.
+ * @param returnValueBytes A buffer containing the return value as raw bytes.
+ * @param moduleSchema The raw module schema as a buffer.
+ * @param contractName The name of the contract where the receive function is located.
+ * @param functionName The name of the receive function whose return value you want to deserialize.
+ * @param schemaVersion The schema version as a number. This parameter is optional,
+ * if you provide a serialized versioned schema this argument won't be needed.
+ */
+export function deserializeReturnValue(
```
shjortConcordium, line 215: Please add an section/example to the common package's README of using this function. (Just as short as it need to be, you can check the deserialize state for reference)
+ Add an entry to the changelog. Just put it under version 5.2.0.

packages/rust-bindings/src/aux_functions.rs
rasmus-kirk: That's the plan, yes. The function is part of [this PR](https://github.com/Concordium/concordium-contracts-common/pull/58)

packages/common/test/deserialization.test.ts
rasmus-kirk: We'll add one for schemaV3 when it comes.

packages/common/test/deserialization.test.ts
```diff
@@ -122,3 +123,17 @@ test('Expired transactions can be deserialized', () => {
         new TransactionExpiry(new Date(2000, 1), true)
     );
 });
+
+test('Receive return value can be deserialized', () => {
```
shjortConcordium, line 127: I made another test (to test this :upside_down_face:), but I won't bother trying to push it to your fork...
```
import * as fs from 'fs';
...
test('Return value can be deserialized - auction', () => {
    const returnValue = deserializeReturnValue(
        Buffer.from(
            '00000b0000004120676f6f64206974656d00a4fbca84010000',
            'hex'
        ),
        Buffer.from(
            fs.readFileSync('./test/resources/auction-with-errors-schema.bin')
        ),
        'auction',
        'view'
    );

    expect(returnValue.item).toEqual('A good item');
    expect(returnValue.end).toEqual('2022-12-01T00:00:00+00:00');
    expect(returnValue.auction_state).toHaveProperty('NotSoldYet');
    expect(returnValue.highest_bidder).toHaveProperty('None');
});

```

Issue #91, opened 2022-10-05
Move serialization of smart contract parameters to rust


shjortConcordium:
**Task description**

We want to remove the duplication of the code for serializing smart contract parameters.

The task would be to replace the implementation of `serializeParameters` in serializationHelpers (in the common-sdk) with a function that calls down into rust-bindings, which would serialize it using concordium-base.

Depending on the format of the schema required for the rust library, this might make `deserialModuleFromBuffer` obsolete.

PR #94, opened 2022-10-18
Moving Parameter serialization to Rust

rasmus-kirk:
## Purpose

Move over the `serializeParameters` functions from NodeJS to Rust, as mentioned in !(this issue)[https://github.com/Concordium/concordium-node-sdk-js/issues/91]:

## Changes

Deleted the functions that serializes parameters from NodeJS and moved them into Rust. Most of the heavy work being done here is by the `packages/rust-bindings/src/schema_json.rs` "stolen" from here, which we would need to import or preferably move.

There is also the `get_receive_schema()` which is duplicated, for now, from this pull request. We could just import this as `concordium-base` is already a dependency, but maybe it should be relocated, preferably along with `get_init_param_schema()` introduced in this commit, to somewhere more general in `concordium-base`?

I am not sure where best to house these, and I am looking for suggestions.


abizjak (changes requested):
Who asked you to do it like this?

We do not copy paste code like this. The correct thing is to move the dependency so it can be used. Copying code does not improve maintainability.

PR #95, opened 2022-10-24
JSON-RPC cookie support

shjortConcordium:
## Purpose

Add cookie support

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/src/providers/httpProvider.ts
+++ b/packages/common/src/providers/httpProvider.ts
@@ -51,8 +51,8 @@ export class HttpProvider implements Provider {
             }
 
             const setCookieValue = res.headers.get('set-cookie');
-            if (onSetCookie && setCookieValue) {
-                onSetCookie(setCookieValue);
+            if (setCookieValue) {
+                onSetCookie?.(setCookieValue);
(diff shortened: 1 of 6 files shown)


packages/common/src/providers/httpProvider.ts
```diff
@@ -48,11 +51,18 @@ export class HttpProvider implements Provider {
             }
 
             const setCookieValue = res.headers.get('set-cookie');
-            if (setCookie && setCookieValue) {
-                setCookie(setCookieValue);
+            if (onSetCookie && setCookieValue) {
+                onSetCookie(setCookieValue);
+                if (autoUpdateCookie) {
+                    this.updateCookie(setCookieValue);
+                }
```
soerenbf, line 58: So all this is supposed to happen only if onSetCookie is set? I would assume the following to be more correct:
```suggestion
            if (setCookieValue) {
                onSetCookie?.(setCookieValue);
                if (autoUpdateCookie) {
                    this.updateCookie(setCookieValue);
                }
```

soerenbf (approved):

PR #96, opened 2022-10-25
only send cookie if set

shjortConcordium:
## Purpose

It gave errors in the browser when sending empty string cookie.

### DIFF
--- a/packages/common/src/providers/httpProvider.ts
+++ b/packages/common/src/providers/httpProvider.ts
@@ -5,7 +5,7 @@ import { v4 as uuidv4 } from 'uuid';
 
 export class HttpProvider implements Provider {
     request: JsonRpcRequest;
-    cookie: string;
+    cookie?: string;
 
     /**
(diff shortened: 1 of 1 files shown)


packages/common/src/providers/httpProvider.ts
```diff
@@ -32,7 +32,7 @@ export class HttpProvider implements Provider {
                 body: JSONBig.stringify(request),
                 headers: {
                     'Content-Type': 'application/json',
-                    cookie: this.cookie,
+                    ...(this.cookie && { cookie: this.cookie }),
```
soerenbf, line 35: This seems weird for me, but it seems to be a legitimate conditional concat 🤷

Would be easier to read like this though IMO:

```ts
const headers = {...};

if (this.cookie !== undefined) {
    headers.cookie = cookie;
}

...
```

If you prefer the current form just ignore this 😉

soerenbf (approved):

PR #111, opened 2022-12-14
Id fixes for country codes and age helpers 

shjortConcordium:
## Purpose

 - Restrict ISO3166_1Alpha2 codes to be upper case.
 - Fix how age helpers handle upper bound of an age.
 - Add explicit checks for each field when verifying statements.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/src/idProofs.ts
+++ b/packages/common/src/idProofs.ts
@@ -164,7 +164,7 @@ function verifySetStatement(
     statement: MembershipStatement | NonMembershipStatement,
     typeName: string
 ) {
-    if (statement.set  === undefined) {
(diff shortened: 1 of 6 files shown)


packages/common/src/idProofs.ts
```diff
@@ -158,6 +164,9 @@ function verifySetStatement(
     statement: MembershipStatement | NonMembershipStatement,
     typeName: string
 ) {
+    if (statement.set  === undefined) {
```
soerenbf, line 167:
```suggestion
    if (statement.set === undefined) {
```

soerenbf (approved):

Issue #102, opened 2022-11-10
Create support for GRPCv2


rasmus-kirk:
**Task Description**
Currently we only have support for GRPCv1 and we need to also support v2.

PR #120, opened 2023-01-25
Grpcv2-PR3: Streams and getblocks

rasmus-kirk:
## Purpose

See #102

## Changes

- Added `getFinalizedBlocks()` & `getBlocks()` GRPCv2 functions.
- Added public helper function `waitForTransactionFinalization()`

### DIFF
--- a/packages/common/src/GRPCClient.ts
+++ b/packages/common/src/GRPCClient.ts
@@ -323,14 +323,15 @@ export default class ConcordiumNodeClient {
     }
 
     /**
-     * Gets a stream of blocks.
+     * Gets a stream of blocks. To get a stream of only finalized blocks
+     * use `getFinalizedBlocks()` instead.
      *
      * @param abortSignal an AbortSignal to close the stream. Note that the
      * stream does not close itself as it is infinite, so usually you'd want
      * to provide this parameter.
-     * @returns An AsyncIterator stream of finalized blocks.
+     * @returns An AsyncIterator stream of blocks.
      */
-    getBlocks(abortSignal?: AbortSignal): AsyncIterable<v1.FinalizedBlockInfo> {
+    getBlocks(abortSignal?: AbortSignal): AsyncIterable<v1.ArrivedBlockInfo> {
         const opts = { abort: abortSignal };
         const blocks = this.client.getBlocks(v2.Empty, opts).responses;
         return mapAsyncIterable(blocks, translate.commonBlockInfo);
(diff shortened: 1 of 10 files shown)


packages/nodejs/test/clientV2.test.ts
```diff
@@ -537,3 +534,40 @@ test.each([clientV2, clientWeb])('createAccount', async (client) => {
         )
     ).rejects.toThrow('expired');
 });
+
+// For tests that take a long time to run, is skipped by default
```
shjortConcordium, line 538:
```suggestion
// Tests, which take a long time to run, are skipped by default
```

packages/common/src/util.ts
```diff
@@ -152,12 +162,27 @@ export function mapRecord<
...
 
-// Retrieves a value that might be undefined. Throws if value is undefined
-export function unwrap<A>(x: A | undefined): A {
-    if (x === undefined) {
-        console.trace();
-        throw Error('Undefined value found.');
-    } else {
-        return x;
-    }
+// Maps an infinite stream of type A to an infinite stream of type B
```
shjortConcordium, line 165:
```suggestion
/** 
 * Maps an infinite stream of type A to an infinite stream of type B
 * @param mapper: function used to map each element from type A to B.   
 */
```

packages/common/src/GRPCClient.ts
```diff
@@ -305,6 +305,63 @@ export default class ConcordiumNodeClient {
...
+
+    /**
+     * Gets a stream of blocks.
+     *
+     * @param abortSignal an AbortSignal to close the stream. Note that the
+     * stream does not close itself as it is infinite, so usually you'd want
+     * to provide this parameter.
+     * @returns An AsyncIterator stream of finalized blocks.
+     */
+    getBlocks(abortSignal?: AbortSignal): AsyncIterable<v1.FinalizedBlockInfo> {
```
shjortConcordium, line 333: I assume these blocks aren't necessarily finalized?
(Change comment / type name?)

packages/common/src/GRPCClient.ts
```diff
@@ -305,6 +305,63 @@ export default class ConcordiumNodeClient {
...
+        return mapAsyncIterable(blocks, translate.commonBlockInfo);
+    }
+
+    /**
+     * Waits until given transaction is finalized.
+     *
+     * @param transactionHash a transaction hash as a bytearray.
+     * @returns A blockhash as a byte array.
+     */
+    async waitForTransactionFinalization(
```
shjortConcordium, line 345: Maybe we should also call `getBlockItemStatus` immediately? :thinking:

shjortConcordium: And/or how do we deal with transactions that don't exist, I assume wait "forever"?

packages/common/src/GRPCClient.ts
rasmus-kirk: Yes

packages/nodejs/src/util.ts
```diff
@@ -1,6 +1,27 @@
...
 import { BoolResponse, JsonResponse } from '../grpc/concordium_p2p_rpc_pb';
+import createConcordiumClientV2 from '../src/clientV2';
+import ConcordiumNodeClientV2 from '@concordium/common-sdk/lib/GRPCClient';
+import { credentials, Metadata } from '@grpc/grpc-js/';
+
+/**
+ * Creates a client to communicate with a local concordium-node
+ * used for automatic tests.
+ */
+export function getNodeClientV2(
```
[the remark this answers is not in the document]
shjortConcordium: (There is a testHelpers file if you just need it for multiple test files)

packages/common/src/GRPCClient.ts
```diff
@@ -305,6 +305,78 @@ export default class ConcordiumNodeClient {
...
+        transactionHash: HexString,
+        timeoutTime?: number
+    ): Promise<HexString> {
+        if (timeoutTime) {
+            setTimeout(() => {
+                throw Error('Function timed out.');
+            }, timeoutTime);
```
shjortConcordium, line 354: This doesn't actually make the function throw. (It throws in a different context)
Also you should use clearTimeout if the the item finalized before the timeout.

packages/common/src/GRPCClient.ts
```diff
@@ -305,6 +305,78 @@ export default class ConcordiumNodeClient {
...
+        }
+
+        const abortController = new AbortController();
+        const blockStream = this.getFinalizedBlocks(abortController.signal);
+
+        const response = await this.getBlockItemStatus(transactionHash);
```
shjortConcordium, line 360: `getBlockItemStatus` throws if the hash does not belong to known transaction? :thinking:

shjortConcordium: It is probably not a problem in actual use cases, but it does make this test unusable:
```
test('waitForTransactionFinalization instant reject', async () => {
    const hash = '3de823b876d05cdd33a311a0f84124079f5f677afb2534c4943f830593edc650';
    await expect(clientV2.waitForTransactionFinalization(
        hash,
        1
    )).rejects.toEqual('Function timed out.')
})
```

PR #132, opened 2023-02-07
Grpcv2-PR9:  `getNodeInfo()`, `getPeersInfo()`

rasmus-kirk:
## Purpose
See #102

## Changes
Added GRPCv2 functions:
 - `getNodeInfo()`
 - `getPeersInfo()`

### DIFF
--- a/packages/common/src/GRPCClient.ts
+++ b/packages/common/src/GRPCClient.ts
@@ -974,7 +974,7 @@ export default class ConcordiumNodeClient {
 
     /**
      * Get a list of the peers that the node is connected to
-     * and assoicated network related information for each peer.
+     * and associated network related information for each peer.
      *
      * @returns a list containing info on each peer of the node.
      */
--- a/packages/common/src/GRPCClient.ts
+++ b/packages/common/src/GRPCClient.ts
@@ -965,11 +965,22 @@ export default class ConcordiumNodeClient {
      * * ConsensusInfo. The `ConsensusInfo` returned depends on if the node supports
      *   the protocol on chain and whether the node is configured as a baker or not.
      *
-     *  @returns Info about the node
+     * @returns Info about the node
      */
     async getNodeInfo(): Promise<v1.NodeInfo> {
         const nodeInfo = await this.client.getNodeInfo(v2.Empty).response;
         return translate.nodeInfo(nodeInfo);
+    }
+
+    /**
+     * Get a list of the peers that the node is connected to
+     * and assoicated network related information for each peer.
+     *
+     * @returns a list containing info on each peer of the node.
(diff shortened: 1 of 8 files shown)


packages/common/src/GRPCClient.ts
```diff
@@ -955,6 +955,33 @@ export default class ConcordiumNodeClient {
...
+     * @returns Info about the node
+     */
+    async getNodeInfo(): Promise<v1.NodeInfo> {
+        const nodeInfo = await this.client.getNodeInfo(v2.Empty).response;
+        return translate.nodeInfo(nodeInfo);
+    }
+
+    /**
+     * Get a list of the peers that the node is connected to
+     * and assoicated network related information for each peer.
```
shjortConcordium, line 977:
```suggestion
     * and associated network related information for each peer.
```

packages/common/src/types/NodeInfo.ts
```diff
@@ -0,0 +1,65 @@
+import { BakerId, Duration, Timestamp } from '../types';
+
+export type PeerId = string;
```
shjortConcordium, line 3:
```suggestion
export type PeerId = HexString;
```
Though it does seem to me the `PeerId` could just be a `HexString`. Like what is a `PeerId`?

packages/nodejs/READMEV2.md
```diff
@@ -834,4 +836,25 @@ Only enabled if the node was built with the `network_dump` feature.
 Rejects if the network dump failed to be stopped.
 ```js
 await this.client.dumpStop();
+```
+
+## getNodeInfo
+Get information about the node.
+
+The `NodeInfo` includes information of:
+ - **Meta information** such as the, version of the node, type of the node, uptime and the local time of the node.
```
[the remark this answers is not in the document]
rasmus-kirk: I don't think so

packages/common/src/types/NodeInfo.ts
```diff
@@ -1,6 +1,4 @@
-import { BakerId, Duration, Timestamp } from '../types';
-
-export type PeerId = HexString;
+import { BakerId, Duration, HexString, Timestamp } from '../types';
```
shjortConcordium, line 1: Wait how did you reference HexString before without importing it :thinking:

shjortConcordium (approved):

packages/common/src/types/NodeInfo.ts
rasmus-kirk: You suggested changing it to a `HexString` and I commited the suggestion in commit `90208bf`, but the suggestion didn't include an import statement. Of course this meant that it could not build too.

PR #149, opened 2023-03-21
Update concordium-base

shjortConcordium:
## Purpose

Update concordium-contracts-common dependency to fix serialization bug.

## Changes

- updated concordium-base submodule

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/src/GRPCTypeTranslation.ts
+++ b/packages/common/src/GRPCTypeTranslation.ts
@@ -1364,6 +1364,9 @@ export function pendingUpdate(
             };
         case undefined:
             throw Error('Unexpected missing pending update');
+        default:
+            // TODO support new updates
+            throw Error('Unsupported update: ' + effect.oneofKind);
     }
 }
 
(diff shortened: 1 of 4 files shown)


packages/common/src/GRPCTypeTranslation.ts
```diff
@@ -1364,6 +1364,9 @@ export function pendingUpdate(
             };
         case undefined:
             throw Error('Unexpected missing pending update');
+        default:
+            // TODO support new updates
+            throw Error('Unsupported update: ' + effect.oneofKind);
```
abizjak, line 1369: Was there an issue where this came up?

abizjak (approved):

packages/common/src/GRPCTypeTranslation.ts
shjortConcordium: Without this, I now get type errors, because I don't handle all possible cases, after the new consensus update types have been added to the gRPC API.

abizjak: thanks.

PR #151, opened 2023-03-22
signMessage and verifyMessageSignature now supports a binary message

shjortConcordium:
## Purpose

Support messages as Buffers (/Uint8Arrays) in signHelper's `signMessage` and `verifyMessageSignature`.

## Changes

Expand the supported message type in `signMessage` and `verifyMessageSignature` to include Uint8Arrays.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/src/signHelpers.ts
+++ b/packages/common/src/signHelpers.ts
@@ -55,24 +55,28 @@ export function signTransaction(
 
 /**
  * @param account the address of the account that will sign this message.
- * @param message the message to sign, assumed to be utf8 encoded.
+ * @param message the message to sign, assumed to be utf8 encoded string or a Uint8Array/buffer.
  */
-function getMessageDigest(account: AccountAddress, message: string): Buffer {
+function getMessageDigest(
+    account: AccountAddress,
+    message: string | Uint8Array
+): Buffer {
     const prepend = Buffer.alloc(8, 0);
-    const rawMessage = Buffer.from(message, 'utf8');
+    const rawMessage =
+        typeof message === 'string' ? Buffer.from(message, 'utf8') : message;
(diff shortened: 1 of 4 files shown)


packages/common/src/signHelpers.ts
```diff
@@ -55,37 +55,41 @@ export function signTransaction(
...
-    const rawMessage = Buffer.from(message, 'utf8');
+    const rawMessage =
+        typeof message === 'string' ? Buffer.from(message, 'utf8') : message;
     return sha256([account.decodedAddress, prepend, rawMessage]);
 }
 
 /**
  * Helper function to sign a message.
- * Note that this function prepends the string "MyGoodPrepend" to ensure that the message is not a transaction.
+ * Note that this function prepends 8 zero-bytes to ensure that the message is not a transaction.
```
abizjak, line 72: I think it technically prepends the sender address (32 bytes) + 8 zero bytes, right?

abizjak (approved):
Looks good, apart from a comment that I think is technically incorrect.

I like the sneaky fix as well :)

If I understand correctly the wallet API will handle the schema and display of the parameter, right? This is just the core SDK function.

shjortConcordium:
> I like the sneaky fix as well :)

Yeah, I ran into it myself this time. :upside_down_face: 

> If I understand correctly the wallet API will handle the schema and display of the parameter, right? This is just the core SDK function.

Yes, exactly, I just wanted to fix it here, instead of doing a weird workaround in the wallet.

packages/common/src/signHelpers.ts
shjortConcordium: Yes, I corrected it. :sweat:

PR #154, opened 2023-03-30
Add examples

rasmus-kirk:
## Purpose
Just a draft for the format of the coming examples


examples/execute-example.bash
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
rasmus-kirk: Maybe we should just use as wrapper JS script though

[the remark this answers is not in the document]
abizjak: Yeah, a lot of people will use Windows, so let's not use bash scripts.

examples/client/getAccountInfo.ts
```diff
@@ -0,0 +1,67 @@
...
+    --account,   -a  An account address to get info from
+    --blockhash, -b  A blockhash to query the info from
+`,
+    {
+        importMeta: import.meta,
+        flags: {
+            endpoint: {
+                type: 'string',
+                shortFlag: 'e',
+                default: 'node.testnet.concordium.com:20000',
```
abizjak, line 24: I think these defaults should be `localhost:20000`.

We don't want people to accidentally spam our node.

examples/client/getAccountInfo.ts
```diff
@@ -0,0 +1,67 @@
...
+            },
+            account: {
+                type: 'string',
+                short: 'a',
+                default: '3kBx2h5Y2veb4hZgAJWPrr8RyQESKm5TjzF3ti1QQ4VSYLwK1G',
```
abizjak, line 29: There shouldn't be default accounts.

examples/client/getAccountInfo.ts
```diff
@@ -0,0 +1,67 @@
...
+            },
+            blockhash: {
+                type: 'string',
+                short: 'b',
+                default:
+                    'fe88ff35454079c3df11d8ae13d5777babd61f28be58494efe51b6593e30716e',
```
abizjak, line 35: We should not have hardcoded blocks like that.

The behaviour that could make sense is to default to "last finalized" if the block is not given.

examples/execute-example.bash
```diff
@@ -0,0 +1,3 @@
...
+path=$(find client common -name "*$1*" | grep '\.ts$' | head)
+yarn ts-node $path "${@:2}"
```
abizjak, line 3: This is incredibly hacky. Why not tell the user to just run

```
yarn ts-node /path/to/example.ts --account --...
```

These sorts of hacky scripts are really not something we should have.

examples/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "eslint-plugin-import": "^2.23.4",
+    "eslint-plugin-prettier": "^3.4.0",
+    "lint-staged": "^12.0.2",
+    "meow": "11.0",
+    "prettier": "^2.3.2",
+    "ts-node": "10.9",
+    "typescript": "^4.3.5"
+  },
+  "scripts": {
+    "example": "./execute-example.bash",
```
shjortConcordium, line 20:
```suggestion
```

And maybe add a readme with how to run the examples. :upside_down_face:

shjortConcordium (commented):
Looks good

examples/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "lint": "eslint . --cache --ext .ts,.tsx --max-warnings 0",
+    "lint-fix": "yarn lint --fix; exit 0",
+    "build": "tsc",
+    "build-dev": "tsc"
```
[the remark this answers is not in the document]
rasmus-kirk: Yes, but won't we need them to check that they typecheck properly? That is the intention of the lines at least.

examples/client/getAccountInfo.ts
```diff
@@ -0,0 +1,68 @@
...
+  Usage
+    $ yarn ts-node <path-to-this-file> [options]
+
+  Required
+    --account,   -a  An account address to get info from
+
+  Options
+    --help,      -h  Displays this message
+    --endpoint,  -e  Specify endpoint of the form "address:port", defaults to localhost
+    --blockhash, -b  A blockhash to query the info from, defaults to last final block
```
abizjak, line 18: It is better to call this just `--block`.

abizjak (approved):

examples/package.json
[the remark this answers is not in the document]
shjortConcordium: Yes, but this way we type check the examples every time we build the SDK's, which I think is wrong to do. 
I think it would make sense to do this in a new CI step instead. :thinking:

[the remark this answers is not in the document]
rasmus-kirk: Yes that makes more sense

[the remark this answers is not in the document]
shjortConcordium: I would call it just `typecheck`or `typecheck:examples` if you need it to be global.

examples/readme.md
```diff
@@ -0,0 +1,12 @@
+## Examples for the Concordium NodeJS-SDK
+
+This is a collection of scripts/examples that utilizes the SDK. There are
+two directories with examples `client`, containing examples that utilize the
+client, and `common`, that use various general functions from the library. To
```
shjortConcordium, line 5:
```suggestion
client to interact with a Concordium node, and `common`, that use various general functions from the library. To
```

Issue #153, opened 2023-03-28
NodeJS-SDK: Runnable examples


rasmus-kirk:
**Sub-tasks**
- [x] `docs/gRPC.md`
    - [x] Create examples 
    - [x] Test examples
    - [x] Pass review
- [x] `packages/common/README.md`
    - [x] Create examples 
    - [x] Test examples
    - [x] Pass review
- [x] Update documentation
    - [x] Setup typedoc
    - [x] Integrate current docs to typedoc
    - [x] Look into inlining examples in docs ([See this](https://github.com/krisztianb/typedoc-plugin-replace-text/issues/4#issuecomment-1549929121))
    - [x] Add a typedoc pipeline

**Example list**
client #164 
  - [x] banPeer.ts
  - [x] dumpStart.ts
  - [x] dumpStop.ts
  - [x] getAccountInfo.ts
  - [x] getAccountList.ts
  - [x] getAccountNonFinalizedTransactions.ts
  - [x] getAncestors.ts
  - [x] getAnonymityRevokers.ts
  - [x] getBakerList.ts
  - [x] getBannedPeers.ts
  - [x] getBlockChainParameters.ts
  - [x] getBlockFinalizationSummary.ts
  - [x] getBlockInfo.ts
  - [x] getBlockItemStatus.ts
  - [x] getBlockPendingUpdates.ts
  - [x] getBlocksAtHeight.ts
  - [x] getBlockSpecialEvents.ts
  - [x] getBlocks.ts
  - [x] getBlockTransactionEvents.ts
  - [x] getBranches.ts
  - [x] getConsensusStatus.ts
  - [x] getCryptographicParameters.ts
  - [x] getElectionInfo.ts
  - [x] getFinalizedBlocks.ts
  - [x] getIdentityProviders.ts
  - [x] getInstanceInfo.ts
  - [x] getInstanceState.ts
  - [x] getModuleList.ts
  - [x] getModuleSource.ts
  - [x] getNextAccountSequenceNumber.ts
  - [x] getNextUpdateSequenceNumbers.ts
  - [x] getNodeInfo.ts
  - [x] getPassiveDelegationInfo.ts
  - [x] getPassiveDelegatorsRewardPeriod.ts
  - [x] getPassiveDelegators.ts
  - [x] getPeersInfo.ts
  - [x] getPoolDelegatorsRewardPeriod.ts
  - [x] getPoolDelegators.ts
  - [x] getPoolInfo.ts
[...]

PR #158, opened 2023-04-13
Runnable Examples: 21 total, added 10

rasmus-kirk:
## Purpose

See #153 

## Changes

Added following runnable examples:

- `getBlockFinalizationSummary.ts`
- `getBlockInfo.ts`
- `getBlockPendingUpdates.ts`
- `getBlockSpecialEvents.ts`
- `getBlockTransactionEvents.ts`
- `getBlocks.ts`
- `getBlocksAtHeightAbsolute.ts`
- `getBlocksAtHeightRelative.ts`
- `getBranches.ts`
- `getConsensusStatus.ts`

## Checklist

      hard-to-understand areas.


examples/client/getBlockFinalizationSummary.ts
```diff
@@ -0,0 +1,59 @@
...
+
+if (cli.flags.h) {
+    cli.showHelp();
+}
+
+/// Get the summary of the finalization data in a given block.
+/// If a blockhash is not supplied it will pick the latest finalized block.
+
+(async () => {
+    const pendingUpdates: AsyncIterable<PendingUpdate> =
```
abizjak, line 50: The name of this file does not fit what is being done here.

examples/client/getBlockInfo.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/// Retrieves information about a specific block.
+/// If a blockhash is not supplied it will pick the latest finalized block.
+
+(async () => {
+    const blockInfo: BlockInfo = await client.getBlockInfo(cli.flags.block);
+
+    console.dir(blockInfo, { depth: null, colors: true });
+
+    // The blockInfo contain information that can then be extracted:
```
abizjak, line 54: We could do a bit more effort, such as printing "a block with N transactions baked at ?. It's parent is ...". Or some such.

examples/client/getBlocks.ts
```diff
@@ -0,0 +1,54 @@
...
+    // Get block stream
+    const blocks: AsyncIterable<ArrivedBlockInfo> = client.getBlocks();
+
+    // Prints blocks infinitely
+    for await (const block of blocks) {
+        console.dir(block, { depth: null, colors: true });
+    }
+
+    // Can also be collected to a list with:
+    const blockList: ArrivedBlockInfo[] = await streamToList(blocks);
```
abizjak, line 53: I'm not sure this will work seeing that the stream never ends.

examples/client/getBlocksAtHeightRelative.ts
```diff
@@ -0,0 +1,72 @@
...
+
+if (cli.flags.h) {
+    cli.showHelp();
+}
+
+/// Get a list of live blocks at a given relative height.
+
+(async () => {
+    const request: BlocksAtHeightRequest = {
+        genesisIndex: 1,
```
abizjak, line 65: This should not be hardcoded. It should also be explained.

examples/client/getBlocksAtHeightRelative.ts
```diff
@@ -0,0 +1,72 @@
...
+        height: BigInt(cli.flags.height),
+        restrict: cli.flags.restrict,
```
abizjak, line 67: Also explain.

examples/client/getBlocksAtHeightRelative.ts
rasmus-kirk: Of course should be `cli.flags.genesisIndex` I'll also flesh out its description on line 16 further. Do you want some explanation here as well.

examples/client/getBlocksAtHeightRelative.ts
rasmus-kirk: Is the explanation on line 14-15 not enough or do you also want a description here.

examples/client/getBlockInfo.ts
rasmus-kirk: I could do a few of those. Do we then want to keep line 52 that prints the whole block then?

The direction I've been going so far is to print the entire returned object, and if applicable, extract information as unused `const`'s that are not printed. Do we want to print things like this or do I continue defining them as unused `const`'s but with better documentation/comments?

examples/client/getBlockFinalizationSummary.ts
```diff
@@ -0,0 +1,64 @@
...
+
+if (cli.flags.h) {
+    cli.showHelp();
+}
+
+/// Get the summary of the finalization data in a given block. Only finalized
+/// blocks will return a finalization summary, if the summary is requested for
+/// a non-finalized block, this will return an object with only the tag field,
+/// with value "none".  If a blockhash is not supplied it will pick the latest
+/// finalized block.
```
soerenbf, line 50: Triple slash comments are not docstrings in typescript (I assume you have this from rust?), but is actually a format reserved for type references. I think we should use proper JS block comments instead

```ts
/**
 * Block comment goes here
 */
```

Same goes for all other files in the PR

soerenbf (changes requested):
I think we should only have references to either `@concordium/node-sdk` or `@concordium/web-sdk` on a per file basis, as developers using the dependency will only have `@concordium/common-sdk` implicitly installed (as a dependency of a dependency)

examples/client/getBlockFinalizationSummary.ts
rasmus-kirk: I knew triple slashes was not "supported" but I still wanted to differentiate between documentation and comments, I did not know triple slashes were reserved however. Using the JS blocks are better.

I'll `sed` it in #164. I'll change all imports to only be `node-sdk` for these client examples in that same PR.

rasmus-kirk: Done now

shjortConcordium (commented):
I agree with @soerenbf 's comment, it makes sense to not use common-sdk directly, because the users won't (/shouldn't) do that.

otherwise LGTM

examples/client/getBlocks.ts
```diff
@@ -0,0 +1,51 @@
...
+        },
+    }
+);
+
+const [address, port] = cli.flags.endpoint.split(':');
+const client = createConcordiumClient(
+    address,
+    Number(port),
+    credentials.createInsecure(),
+    { timeout: 15000 }
```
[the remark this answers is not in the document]
soerenbf: Yeah.. people will ctrl-C anyhow if it runs for too long 😄

soerenbf (approved):
LGTM by now and with the promise of including implementations of changes in resolved comments in subsequent PR's.

I think, however, we should limit this workflow (fixing requested changes for one PR in another one) to the absolute minimum, as it in essence circumvents the purpose of approving changes to be included in the target branch, when in fact the changes are merely promised to be implemented at a later point in time.

If you want someone else to also approve, wait with merging until they've had a chance to look over the changes as well.

examples/client/getBlockInfo.ts
abizjak: It's up to you, but I would go for something like printing a few interesting things only. Dumping large objects is not super nice. They should get intellisense support in VSCode I assume when modifying the example.

examples/client/getBlocksAtHeightRelative.ts
abizjak: yes. Always :)

abizjak:
I support @soerenbf in expressing that it is almost never OK to say "I'll fix this in a followup PR". Structure PRs in a way that that is not necessary.

examples/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "@typescript-eslint/no-unused-vars": 0
```
abizjak, line 3: I don't think you should do this. Print them instead./

abizjak (approved):
Address the existing comments. Otherwise looks ok.

rasmus-kirk:
All right I addressed the comments and some other issues I stumbled on: 
- Printed more information wherever I saw it fit
- Removed the `-h` flag, it had inconsistent behavior with `--help`
- Removed all unused `const`'s
- Don't do `streamToList` on stream examples, but mention it in the documentation.
- I don't import from `common` anymore, only `nodejs`
- Documentation comments now have correct style

shjortConcordium (approved):

PR #161, opened 2023-04-17
Runnable examples: 31 total, 10 added

rasmus-kirk:
## Purpose

See #153 

## Changes

Added following examples
 - `getCryptographicParameters`
 - `getElectionInfo`
 - `getFinalizedBlocks`
 - `getIdentityProviders`
 - `getInstanceInfo`
 - `getInstanceState`
 - `getModuleList`
 - `getModuleSource`
 - `getNextAccouequenceNumber`
 - `getNextUpdateSequenceNumbers`

## Checklist

      hard-to-understand areas.


examples/client/getFinalizedBlocks.ts
shjortConcordium, line 1: I think this example should work the same way as the `getBlocks`.  (or `getBlocks` should work like this)

examples/client/getInstanceInfo.ts
```diff
@@ -0,0 +1,72 @@
...
+    };
+
+    const instanceInfo: InstanceInfo = await client.getInstanceInfo(
+        contractAddress,
+        cli.flags.block
+    );
+
+    console.dir(instanceInfo, { depth: null, colors: true });
+
+    // The instanceInfo contain information that can then be extracted:
```
shjortConcordium, line 70:
```suggestion
    // The instanceInfo contains information that can then be extracted:
```

examples/client/getElectionInfo.ts
```diff
@@ -0,0 +1,58 @@
...
+/// If a blockhash is not supplied it will pick the latest finalized block.
+
+(async () => {
+    const electionInfo: ElectionInfo = await client.getElectionInfo(
+        cli.flags.block
+    );
+
+    console.dir(electionInfo, { depth: null, colors: true });
+
+    // The electionInfo contain information that can then be extracted:
```
shjortConcordium, line 56:
```suggestion
    // The electionInfo contains information that can then be extracted:
```

examples/client/getModuleSource.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/// Gets the source of a module on the chain.
+/// Note that this returns the raw bytes of the source, as a HexString.
+
+(async () => {
+    const ref = new ModuleReference(cli.flags.module);
+    const source = await client.getModuleSource(ref, cli.flags.block);
+
+    console.log(source);
```
shjortConcordium, line 61: This just prints the beginning of the source:
```
<Buffer 00 61 73 6d 01 00 00 00 01 a8 01 19 60 02 7f 7f 00 60 04 7f 7f 7f 7f 00 60 02 7f 7f 01 7f 60 03 7f 7f 7f 01 7f 60 01 7e 01 7f 60 03 7f 7f 7f 01 7e 60 ... >
```

Maybe we can display it better. (or maybe we don't want to print the entire blob? :thinking: )

shjortConcordium (commented):
Generally LGTM
(Importing from `@concordium/common-sdk` applies here too though)

examples/client/getModuleSource.ts
rasmus-kirk: Wait why does this even return a buffer when everything else returns a `HexString`? We should definitely display it better though, I'll make it print the blob as a hex string.

rasmus-kirk: I made it write to a file instead

examples/client/getFinalizedBlocks.ts
shjortConcordium: It still times out now (like getBlocks :thinking: )
I do like the new example though, just not the name (I think I prefer getOneBlock)

rasmus-kirk: Ah yes, the timeout will be fixed in #164 on all examples, I'll just remove it as discussed previously. The name is so that it's in alphabetic order. Would `getBlockSingle` or `getBlockOne` be alright? Otherwise I'll just rename it to `getOneBlock`

examples/client/getModuleSource.ts
shjortConcordium: Should we? :thinking:  :shrug:

examples/client/getFinalizedBlocks.ts
shjortConcordium: I think `getBlockSingle` is okay. Is alphabetic order important? :thinking:

examples/client/getModuleSource.ts
rasmus-kirk: Seems less simple to write to a file, but getting a very long hex/base64 output also seems wrong. To me it seemed least bad, but I really don't have strong opinions on it.

examples/client/getFinalizedBlocks.ts
rasmus-kirk: Not really, but I think that it's nice that they are in order when the tests are so similar

examples/client/getModuleSource.ts
rasmus-kirk: Ah makes sense, I mean it's not ideal that it's a buffer when everything else is hex strings, right? On the other hand it's also bad to break compatibility.

It doesn't matter much though.

soerenbf (approved):
Looks good :)

shjortConcordium (approved):

PR #164, opened 2023-04-19
Runnable examples: 47 total, 7 added

rasmus-kirk:
## Purpose

See #153 

## Changes

Added examples:

 - `getTokenomicsInfo`
 - `instanceStateLookup`
 - `invokeContract`
 - `peerConnect`
 - `peerDisconnect`
 - `shutdown`
 - `unbanPeer`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/client/banPeer.ts
+++ b/examples/client/banPeer.ts
@@ -9,7 +9,7 @@ const cli = meow(
     $ yarn run-example <path-to-this-file> [options]
 
   Required
-    --peer, -p  The ip of the peer to ban.
+    --ip, -i  The ip of the peer to ban.
 
   Options
     --help,         Displays this message
@@ -18,9 +18,9 @@ const cli = meow(
     {
         importMeta: import.meta,
         flags: {
-            peer: {
+            ip: {
                 type: 'string',
-                alias: 'p',
+                alias: 'i',
                 isRequired: true,
             },
             endpoint: {
@@ -45,6 +45,6 @@ const client = createConcordiumClient(
  */
 
 (async () => {
-    await client.banPeer(cli.flags.peer);
+    await client.banPeer(cli.flags.ip);
     console.log('Specified peer successfully banned');
 })();
--- a/examples/client/banPeer.ts
+++ b/examples/client/banPeer.ts
@@ -1,3 +1,4 @@
+import { parseEndpoint } from '../shared/util';
 import { createConcordiumClient } from '@concordium/node-sdk';
 import { credentials } from '@grpc/grpc-js';
 
@@ -32,10 +33,7 @@ const cli = meow(
     }
 );
 
-// Split endpoint on last colon
-const lastColonIndex = cli.flags.endpoint.lastIndexOf(':');
-const address = cli.flags.endpoint.substring(0, lastColonIndex);
-const port = cli.flags.endpoint.substring(lastColonIndex + 1);
+const [address, port] = parseEndpoint(cli.flags.endpoint);
 
 const client = createConcordiumClient(
     address,
--- a/examples/client/banPeer.ts
+++ b/examples/client/banPeer.ts
@@ -32,7 +32,11 @@ const cli = meow(
     }
 );
 
-const [address, port] = cli.flags.endpoint.split(':');
+// Split endpoint on last colon
+const lastColonIndex = cli.flags.endpoint.lastIndexOf(':');
+const address = cli.flags.endpoint.substring(0, lastColonIndex);
+const port = cli.flags.endpoint.substring(lastColonIndex + 1);
+
 const client = createConcordiumClient(
     address,
     Number(port),
--- a/examples/client/instanceStateLookup.ts
+++ b/examples/client/instanceStateLookup.ts
@@ -37,7 +37,6 @@ const cli = meow(
             block: {
                 type: 'string',
                 alias: 'b',
-                default: '', // This defaults to LastFinal
             },
             endpoint: {
                 type: 'string',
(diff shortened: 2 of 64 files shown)


examples/client/banPeer.ts
```diff
@@ -9,10 +9,10 @@ const cli = meow(
     $ yarn ts-node <path-to-this-file> [options]
 
   Required
-    --peer, -p  The ip of the peer to ban.
+    --ip, -i  The ip of the peer to ban.
 
   Options
-    --help,     -h  Displays this message
+    --help,         Displays this message
```
rasmus-kirk, line 15: Removed `-h` alias as it had inconsistent behavior with `--help`

examples/client/banPeer.ts
```diff
@@ -40,14 +40,12 @@ const client = createConcordiumClient(
     { timeout: 15000 }
 );
 
-if (cli.flags.h) {
-    cli.showHelp();
-}
```
rasmus-kirk, line 45: Typing `--help` will still work as expected.

examples/client/invokeContract.ts
```diff
@@ -0,0 +1,140 @@
...
+            },
+            amount: {
+                type: 'number',
+                alias: 'a',
+                default: NaN,
+            },
+            energy: {
+                type: 'number',
+                alias: 'n',
+                default: NaN,
```
soerenbf, line 58: I don't think `NaN` would make any sense here?

```suggestion
                default: 10000,
```

examples/client/invokeContract.ts
```diff
@@ -0,0 +1,140 @@
...
+            },
+            block: {
+                type: 'string',
+                alias: 'b',
+                default: '', // This defaults to LastFinal
+            },
+            amount: {
+                type: 'number',
+                alias: 'a',
+                default: NaN,
```
soerenbf, line 53:
```suggestion
                default: 0,
```

examples/client/invokeContract.ts
```diff
@@ -0,0 +1,140 @@
...
+            },
+            contract: {
+                type: 'number',
+                alias: 'c',
+                isRequired: true,
+            },
+            block: {
+                type: 'string',
+                alias: 'b',
+                default: '', // This defaults to LastFinal
```
soerenbf, line 48:
```suggestion
                default: '', // This defaults to LastFinal
```

Makes it undefined instead, which should have the same result, just cleaner in my opinion

soerenbf, line 48:
```suggestion
```

Makes it undefined instead, which should have the same result, just cleaner in my opinion

examples/client/invokeContract.ts
```diff
@@ -0,0 +1,140 @@
...
+            },
+            endpoint: {
+                type: 'string',
+                alias: 'e',
+                default: 'localhost:20000',
+            },
+            parameter: {
+                type: 'string',
+                alias: 'p',
+                default: '',
```
soerenbf, line 73:
```suggestion
```

examples/client/invokeContract.ts
```diff
@@ -0,0 +1,140 @@
...
+            },
+            energy: {
+                type: 'number',
+                alias: 'n',
+                default: NaN,
+            },
+            invoker: {
+                type: 'string',
+                alias: 'i',
+                default: '',
```
soerenbf, line 63:
```suggestion
```

examples/client/instanceStateLookup.ts
```diff
@@ -0,0 +1,78 @@
...
+            },
+            contract: {
+                type: 'number',
+                alias: 'c',
+                isRequired: true,
+            },
+            block: {
+                type: 'string',
+                alias: 'b',
+                default: '', // This defaults to LastFinal
```
soerenbf, line 40:
```suggestion
```

examples/client/invokeContract.ts
rasmus-kirk: `NaN` behaves the same as `undefined`, so that the `invokeContract` call will choose the default value itself, which is cleaner than an arbitrary value for this example.

It should be `undefined` instead of `NaN` though, so I'll remove the default value entirely here if that works for you.

examples/client/invokeContract.ts
rasmus-kirk: I agree, I'll go ahead and do it in the other examples as well

rasmus-kirk:
@shjortConcordium Addressed your comments, also fixed the issue with the endpoint colons we talked about

examples/client/banPeer.ts
```diff
@@ -32,7 +32,11 @@ const cli = meow(
     }
 );
 
-const [address, port] = cli.flags.endpoint.split(':');
+// Split endpoint on last colon
+const lastColonIndex = cli.flags.endpoint.lastIndexOf(':');
+const address = cli.flags.endpoint.substring(0, lastColonIndex);
+const port = cli.flags.endpoint.substring(lastColonIndex + 1);
+
```
soerenbf, line 39: This is pretty much in every file. We should have a helper function to parse this instead, so we could have something like

```ts
const { address, port } = parseEndpoint(cli.flags.endpoint);
```

Note that that changes how the example should be run to `yarn run-example <path-to-script> [options]`, as otherwise loading local modules does not work, so we should update the description of how to run the script to match this.

I know this is annoying to have to change everywhere... Sorry! 🙈

soerenbf (approved):

examples/client/banPeer.ts
rasmus-kirk: Shouldn't take too long

> _`sed` giveth, `sed` taketh away_

soerenbf (approved):

shjortConcordium (approved):

PR #167, opened 2023-04-27
Added getEmbeddedSchema, updated `common` dependency and small fixups

rasmus-kirk:
## Purpose & Changes

- Added `getEmbeddedSchema` that I needed for some examples anyways
- Updated concordium-nodejs's dependency of concordium-common to `6.4.2`
- Some small cleanups I cought in `clientV2.test.ts`

## Checklist

      hard-to-understand areas.


packages/common/src/util.ts
```diff
@@ -137,6 +138,26 @@ export function countSignatures(
     return totalSignatureCount;
 }
 
+/**
+ * Converts a wasm module to a smart contract schema.
+ * @param wasm the wasm module as a Buffer
+ * @returns the smart contract schema as a Buffer
+ */
+export function wasmToSchema(wasm: Buffer): Buffer {
```
shjortConcordium, line 146: I think it could make sense to test this and check the "correct" schema is extracted.
I.e. without the node, where we have an expected schema to check against.

otherwise LGTM.

rasmus-kirk: Good suggestion

rasmus-kirk: Done, take a look. Note that I found a bug in `getModuleSource` while working on this: #173

shjortConcordium (approved):

PR #178, opened 2023-05-04
Exposed RpcError and created isRpcError function

rasmus-kirk:
## Purpose

Allows users to check errors without resorting to typecasting them to the `any` type. For example:

```ts
try {
    await client.getAccountInfo(accountAddress);
} catch (e) {
    if (isRpcError(e) && e.code === 'NOT_FOUND') {
        // Handle error here
    } else {
        throw e
    }
}
```

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -9,7 +9,8 @@
 - Made the `streamToList` function public.
 - Made the `unwrap` function public.
 - Added `wasmToSchema` utility function.
-- Added `getEmbeddedSchema` to client.>>>>>>> main
+- Added `getEmbeddedSchema` to client.
+- Exposed `RpcError` type and created helper `isRpcError`.
 - Build function `buildAccountSigner` for creating `AccountSigner` objects from genesis format, wallet export format, and a simple representation of credentials with keys.
 
 ## 6.4.2 2023-04-21
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -13,6 +13,9 @@
 - Exposed `RpcError` type and created helper `isRpcError`.
 - Build function `buildAccountSigner` for creating `AccountSigner` objects from genesis format, wallet export format, and a simple representation of credentials with keys.
 
+### Changed
+- Fixed bug where `AccountCreationSummary` type did not have fields: `index`, `energyCost`, `hash`
+
 ## 6.4.2 2023-04-21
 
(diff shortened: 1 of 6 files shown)


packages/common/CHANGELOG.md
```diff
@@ -9,7 +9,8 @@
 - Made the `streamToList` function public.
 - Made the `unwrap` function public.
 - Added `wasmToSchema` utility function.
-- Added `getEmbeddedSchema` to client.>>>>>>> main
+- Added `getEmbeddedSchema` to client.
```
shjortConcordium, line 12: :upside_down_face:

packages/common/src/types/errors.ts
```diff
@@ -0,0 +1,7 @@
+import { RpcError } from '@protobuf-ts/runtime-rpc';
+
+export { RpcError };
+
+export function isRpcError(error: unknown): error is RpcError {
+    return (error as RpcError).code !== undefined;
+}
```
shjortConcordium, line 7: I think we should check it a little more thorough here :thinking: 
Looking at the type is also always has a meta field and a name field (which seems to always be "RpcError")

shjortConcordium (commented):
I think it would make sense to add, basically, the issue description as a section in the gRPC doc.

packages/common/src/types/errors.ts
rasmus-kirk: I agree, I'll check all other required fields as well

rasmus-kirk: And add it to the documentation

shjortConcordium (approved):

packages/common/src/types/errors.ts
```diff
@@ -0,0 +1,13 @@
...
+export { RpcError };
+
+export function isRpcError(error: unknown): error is RpcError {
+    const err = error as RpcError;
+    return (
+        err.code !== undefined &&
+        err.meta !== undefined &&
+        err.name !== undefined &&
+        err.toString !== undefined
+    );
```
soerenbf, line 12:
```suggestion
    return error instanceof RpcError;
```

Since `RpcError` is a class, this should also work, and would be more precise as well

rasmus-kirk: Much better, I'll test it and merge it, okay?

soerenbf: Sure 😄

Issue #166, opened 2023-04-27
Return `BlockItemSummary` directly from `ConcordiumNodeClient.waitForTransactionFinalization`


soerenbf:
In the `ConcordiumNodeClient` class, it would be nice if the following alteration was made, to align the interface with the GRPCv2 client counterpart in the rust SDK:

**waitForTransactionFinalization**
This method currently returns only the block hash of the block the transaction is included in. It would be useful if the `BlockItemSummary` for the finalized transaction would be returned in addition to this, instead of having to issue another GRPC request for this.

See [v2::client.wait_until_finalized](https://docs.rs/concordium-rust-sdk/latest/concordium_rust_sdk/v2/struct.Client.html#method.wait_until_finalized) for reference

Issue #112, opened 2022-12-20
Make easier to use with server-side-rendering frameworks


soerenbf:
**Bug Description**

Currently, using `@concordium/web-sdk` in an app that does server side rendering (SSR) causes the server doing SSR to crash due to not being able to find a suitable module.

The current workaround woud be to resolve `@concordium/web-sdk` to `@concordium/node-sdk`. This is going to be a challenge for a lot of app developers, as it is pretty normal to use frameworks such as nextJS, nuxt, etc. to develop web apps.

Currently, we specify only `browser` and `types` field in the published `@concordium/web-sdk/package.json`.

A solution could be to simply include the node-sdk as part of the web-sdk:

file tree:
```
/
  package.json
  lib/
    concordium.min.js <-- browser code
    index.js <-- node-js code
    index.d.ts <-- types
```

package.json:
```json
{
  "main": "lib/index.js"
  "types": "lib/index.d.ts"
  "browser": "lib/concordium.min.js"
  "files": ["lib/*"]
}
```

We would have to ensure, however, that the nodejs-sdk is identical or a superset of the web-sdk for this to work consistently.

Adding this should not be a breaking change.

PR #180, opened 2023-05-08
Fixed bug where `getBlockInfo` throws error on genesis blocks.

rasmus-kirk:
## Changes

- Updated `blockInfo` so that the `bakerId` field is optional, since it will be undefined for genesis blocks.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-### Changed
+### Breaking changes
 - Updated `blockInfo` so that the `bakerId` field is optional, since it will be undefined for genesis blocks.
 
 ## 6.5.0 2023-5-03
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Changed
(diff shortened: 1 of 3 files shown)


rasmus-kirk:
This is a breaking change, and will not be merged until the next major release.

abizjak:
I am not sure about the rationale here. This is a bug where the SDK does not work. We can be academic about it and delay the release because it is a "major version bump", but ultimately the SDK in the current version is broken.

How long are we waiting with this release then?

soerenbf:
> I am not sure about the rationale here. This is a bug where the SDK does not work. We can be academic about it and delay the release because it is a "major version bump", but ultimately the SDK in the current version is broken.
> 
> How long are we waiting with this release then?

Well we should not wait too long then. Would be nice to at least include #166 and possibly check if we can implement a good solution for #112 (I don't know if this would actually require anything breaking the current implementation or not, but we could at least investigate).

packages/common/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Changed
```
shjortConcordium, line 5: The section should be breaking changes. :thinking:

shjortConcordium (approved):

PR #185, opened 2023-05-10
Updated CI

rasmus-kirk:
## Purpose

Update the CI, with caching, typechecking of examples and general cleanup

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -38,7 +38,7 @@ jobs:
             ./packages/rust-bindings/pkg
             ./packages/common/lib
             ./packages/nodejs/lib
-          key: ${{ runner.os }}-build_output-${{ hashFiles('**/*.ts', '**/yarn.lock', '.git/modules/deps/concordium-base/HEAD', './packages/rust-bindings/Cargo.lock', '.packages/rust-bindings/src/*.rs') }}
+          key: ${{ runner.os }}-build_output-${{ hashFiles('.git/refs') }}
 
       - name: Cache GRPC
         id: cache-grpc
@@ -97,7 +97,7 @@ jobs:
             ./packages/rust-bindings/pkg
             ./packages/common/lib
             ./packages/nodejs/lib
-          key: ${{ runner.os }}-build_output-${{ hashFiles('**/*.ts', '**/yarn.lock', '.git/modules/deps/concordium-base/HEAD', './packages/rust-bindings/Cargo.lock', '.packages/rust-bindings/src/*.rs') }}
+          key: ${{ runner.os }}-build_output-${{ hashFiles('.git/refs') }}
 
       - name: Cache GRPC
         id: cache-grpc
@@ -158,7 +158,7 @@ jobs:
             ./packages/rust-bindings/pkg
             ./packages/common/lib
             ./packages/nodejs/lib
-          key: ${{ runner.os }}-build_output-${{ hashFiles('**/*.ts', '**/yarn.lock', '.git/modules/deps/concordium-base/HEAD', './packages/rust-bindings/Cargo.lock', '.packages/rust-bindings/src/*.rs') }}
+          key: ${{ runner.os }}-build_output-${{ hashFiles('.git/refs') }}
 
       - name: Cache GRPC
         id: cache-grpc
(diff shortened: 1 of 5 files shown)


.github/workflows/pipeline.yml
```diff
@@ -1,136 +1,220 @@
...
-
+        # Don't run on draft PR's, see: https://github.com/orgs/community/discussions/25722#discussioncomment-3248917
+        types: [opened, synchronize, reopened, ready_for_review]
     # Allows us to run the workflow manually from the Actions tab
     workflow_dispatch:
 
 env:
+  NODE_VERSION: 18.16.0
+  RUST_VERSION: 1.62
   RUST_FMT: nightly-2022-06-09-x86_64-unknown-linux-gnu
```
soerenbf, line 17: We should update this to match the formatter we changed the rest of our projects to run with:
```suggestion
  RUST_FMT: nightly-2023-04-01-x86_64-unknown-linux-gnu
```

.github/workflows/pipeline.yml
```diff
@@ -1,136 +1,220 @@
...
         branches: [main, release**]
-
+        # Don't run on draft PR's, see: https://github.com/orgs/community/discussions/25722#discussioncomment-3248917
+        types: [opened, synchronize, reopened, ready_for_review]
     # Allows us to run the workflow manually from the Actions tab
     workflow_dispatch:
 
 env:
+  NODE_VERSION: 18.16.0
+  RUST_VERSION: 1.62
```
soerenbf, line 16: Seems we use this in other projects:
```suggestion
  RUST_VERSION: 1.68
```

.github/workflows/pipeline.yml
rasmus-kirk: I agree but then we need to bump concordium-base as well, which is not really a problem, I'll just do that.

soerenbf: I think we should :)

packages/rust-bindings/package.json
```diff
@@ -17,8 +17,6 @@
         "/pkg/**/concordium_rust_bindings*"
     ],
     "scripts": {
-        "lint": "cargo clippy",
-        "lint-fix": "cargo +nightly-2022-06-09-x86_64-unknown-linux-gnu fmt",
```
soerenbf, line 21: Why remove these? Wouldn't these still be useful to apply the formatting expected by the CI flow?

Also, align the version used for fmt with the new version in the CI (if we still need these scripts).

shjortConcordium: They already have separate jobs in the CI
(The version used in the fmt job should probably be updated though, to nightly-2023-04-01, can you do that here @rasmus-kirk :pleading_face: )

rasmus-kirk: Yeah, I'll do that

rasmus-kirk: Done, take a look

soerenbf (approved):

.github/workflows/pipeline.yml
```diff
@@ -38,7 +38,7 @@ jobs:
             ./packages/rust-bindings/pkg
             ./packages/common/lib
             ./packages/nodejs/lib
-          key: ${{ runner.os }}-build_output-${{ hashFiles('**/*.ts', '**/yarn.lock', '.git/modules/deps/concordium-base/HEAD', './packages/rust-bindings/Cargo.lock', '.packages/rust-bindings/src/*.rs') }}
+          key: ${{ runner.os }}-build_output-${{ hashFiles('.git/refs') }}
```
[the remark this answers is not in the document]
abizjak: Are you sure this is the same `/refs` contains a whole bunch of crap. Typically here we want to explicitly list the files that have changed, to be explicit about it.

Using `.git/refs` seems to, if it works, to work by accident.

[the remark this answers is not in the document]
rasmus-kirk: For the build-output specifically, we would want to only cache it for jobs that depend on it, because any meaningful change you make to the SDK would require a rebuild anyways, so the question to ask is: "Are we on a unique commit?". 

This means I essentially just wanted the HEAD of the current commit ref. I have later found out that we can do `env.GIT_SHA` which gives me what I originally wanted.

PR #191, opened 2023-05-15
Name gRPC client consistently

bisgardo:
## Purpose

Name gRPC client consistently as I find it confusing that it's known under both `ConcordiumGRPCClient` and `ConcordiumNodeClient`.

## Changes

Rename `ConcordiumNodeClient` to `ConcordiumGRPCClient` as that's the name under which it's exported. It looks to me like `ConcordiumNodeClient` is used only internally within common-sdk.

Also changed the export from `default` to named by the class.

If it cannot be renamed for some reason that I missed, I still think the README should be updated to reflect the exported name.

## Checklist

      hard-to-understand areas.


shjortConcordium (commented):
LGTM, there were some places where you missed, so I've added them.
I don't think this would have any effects externally (outside of documentation), @orhoj?

orhoj:
Looks good to me, but the CHANGELOG has to be updated to reflect the changes made.

bisgardo:
> Looks good to me, but the CHANGELOG has to be updated to reflect the changes made.

Are you sure? I believe the changes are all internal (except for the documentation changes which I consider fixes but not related to a particular version). I think changelog is for externally visible changes.

bisgardo:
> > Looks good to me, but the CHANGELOG has to be updated to reflect the changes made.
> 
> Are you sure? I believe the changes are all internal (except for the documentation changes which I consider fixes but not related to a particular version). I think changelog is for externally visible changes.

ping @orhoj

soerenbf:
> > Looks good to me, but the CHANGELOG has to be updated to reflect the changes made.
> 
> Are you sure? I believe the changes are all internal (except for the documentation changes which I consider fixes but not related to a particular version). I think changelog is for externally visible changes.

`@concordium/common-sdk` is a published library (albeit not directly used by end users), but the breaking changes should still be documented in the CHANGELOG.

Theres a release branch collecting the next round of breaking changes (PR #206), which you could set as the base branch for this PR.

bisgardo:
> > > Looks good to me, but the CHANGELOG has to be updated to reflect the changes made.
> > 
> > 
> > Are you sure? I believe the changes are all internal (except for the documentation changes which I consider fixes but not related to a particular version). I think changelog is for externally visible changes.
> 
> `@concordium/common-sdk` is a published library (albeit not directly used by end users), but the breaking changes should still be documented in the CHANGELOG.
> 
> Theres a release branch collecting the next round of breaking changes (PR #206), which you could set as the base branch for this PR.

But my point is that the changes are not breaking because [the exported name is unchanged](https://github.com/Concordium/concordium-node-sdk-js/pull/191/files#diff-970bb4d849f8275a7e4a4d17f71211e3a99fac7e27a9f3c0b826746287135ddbR50). Or am I missing something?

soerenbf:
> But my point is that the changes are not breaking because [the exported name is unchanged](https://github.com/Concordium/concordium-node-sdk-js/pull/191/files#diff-970bb4d849f8275a7e4a4d17f71211e3a99fac7e27a9f3c0b826746287135ddbR50). They're not externally visible at all. Or am I missing something?

It is public: https://www.npmjs.com/package/@concordium/common-sdk

bisgardo:
> I see what you mean, but you can theoretically still import the class from the actual file `lib/GRPCClient` instead of from the index file, in which case it would be breaking? Maybe it doesn't matter when it is not the export from index, I honestly have no idea.
> 
> Anyhow, if we include this in the next major release it won't matter anyway?

I suppose that would break (I expected the whole lib to be compiled down to a single file). I think whether or not it's a breaking change then boils down to whether or not this is a supported use case? Don't we want users to import everything from `@concordium/common-sdk`?

I can fix the merge conflicts (it's just a string search/replace) and also add the changelog entry if you think it's relevant (I still don't think it is)?

soerenbf:
> I suppose that would break (I expected the whole lib to be compiled down to a single file). I think whether or not it's a breaking change then boils down to whether or not this is a supported use case? Don't we want users to import everything from `@concordium/common-sdk`?
> 
> I can fix the merge conflicts (it's just a string search/replace) and also add the changelog entry if you think it's relevant (I still don't think it is)?

Yeah, let's just skip the changelog as the top level stays the same.. I think that's fair.

soerenbf (approved):

bisgardo:
@soerenbf Rebased and included a few more replacements (see https://github.com/Concordium/concordium-node-sdk-js/pull/191/commits/83534b1f41cfa1723ded82777772e7cb588df80a).

I'm assuming that the remaining occurrences of `ConcordiumNode` (in packages/nodejs/src/client.ts, packages/nodejs/src/index.ts, and pages/misc-pages/grpc-v1.md) are for V1 and shouldn't be replaced?

soerenbf (approved):
My only concern was the generated docs, but they seem to look good and work as expected!

PR #195, opened 2023-05-19
Fix a few error messages

bisgardo:
In what looks like copy/paste errors, two "serialize" functions return errors saying that "deserialization" failed.

While at it, I improved consistency of the other messages in the file.


DOBEN (approved):
LGTM, the linter is complaining.

rasmus-kirk (approved):

PR #199, opened 2023-06-01
Improve (de)serialization errors

### LINKED ISSUE #187: Improved error reporting when serializing/deserializing in node JS SDK
`Concordium-client` can give good error messages when part of the e.g. input parameter is wrongly provided (e.g. `Error message: In the JSON input parameter in field `token_ID` a `hex string` is expected but got `xyz` instead`). We should have similar error messages pointing out exactly the field/place that is wrong when using the node SDK functions
- `signAndSendTransaction`
- `serializeUpdateContractParameters`
- `deserializeReceiveReturnValue`

Currently, theses three functions just not serialize/deserialize the value if something is wrong but it is difficult to pinpoint the exact field/area that was wrongly provided. Not exactly sure if `deserializeReceiveReturnValue` falls under this issue as well but I just mentioned it to investigate.

### LINKED ISSUE #188: Update JSON parameter serialization error to include trace from object root to failure point
**Description**
To allow developers to easily identify why serialization of JSON formatted contract parameters fails, we need to add trace information to the location of the serialization failure in the object.

### LINKED ISSUE #190: Update deserialization functions to indicate the origin of failures
**Description**
While serialization of return values, events, and errors in smart contracts is not manual (like when developers have to structure JSON parameters to be compatible with the contract schema), it might still provide some form of value to be able to pinpoint why deserialization of these fails. As of now, many of these don’t provide any detail of what fails and where. To help developers identify where the error happens, it would make sense to provide point in the buffer at which deserialization failed, and which type it could not be deserialized to.

### LINKED ISSUE #196: `Try .. catch` block is missing in `deserializeTypeValue` function which causes the errors not to be thrown correctly
**Bug Description**

`Try .. catch` block is missing in `deserializeTypeValue` function which causes the errors not to be thrown correctly.
https://github.com/Concordium/concordium-node-sdk-js/blob/main/packages/common/src/deserialization.ts#L321

See the `Try .. catch` blocks in the above functions on how errors are thrown.

**Steps to Reproduce**

Running the [front-end] (https://github.com/Concordium/concordium-misc-tools/pull/55) and choosing the `wrong_schema` option at one of the two `dropDowns` at the front end then pressing the button in that test unit should cause an error. When comparing these two errors:
 
When comparing the errors returned by the functions `deserializeReceiveReturnValue()` and `deserializeTypeValue()`:

The error from `deserializeTypeValue()` can not be cast correctly. `(err as string)` is used in the below line to overcome this issue:
https://github.com/Concordium/concordium-misc-tools/blob/171c7a82549db58f1db5d53b21ec1a7805754315/wallet-connect-test-bench/src/Transactions.tsx#L170

The error from `deserializeReceiveReturnValue()` can be cast correctly. `(err as Error).message` is used in the below line:
https://github.com/Concordium/concordium-misc-tools/blob/171c7a82549db58f1db5d53b21ec1a7805754315/wallet-connect-test-bench/src/Transactions.tsx#L158 (edited)

soerenbf:
## Purpose

Improves errror messages when serializing and deserializing using contract schemas. Most of the work done lies in `concordium-contracts-common` and was merged with https://github.com/Concordium/concordium-contracts-common/pull/87. This updates the wasm API to match the added functionality.

Furthermore, errors returned from wasm are now properly formatted JS errors. We need to decide if this is a breaking change or not, as the type of the error is infact changed from `string` to `Error`.

Closes #187, closes #188, closes #190, closes #196

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/rust-bindings/CHANGELOG.md
+++ b/packages/rust-bindings/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Breaking changes
+
+- Errors returned from entrypoints in `external_functions.rs` are now proper JS errors.
+
+### Changes
+
+- Added an additional parameter to contract schema serialization/deserializtion entrypoints, making it possible to receive errors in a verbose format with added detail.
+
 ## 0.12.0
 
 ### Added
--- a/packages/rust-bindings/CHANGELOG.md
+++ b/packages/rust-bindings/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.12.0
+
+### Added
+
+- Methods for deriving verifiable credentials keys from a seed phrase.
+
 ## 0.11.1 2023-4-21
 
 ### Changes
--- a/packages/rust-bindings/src/external_functions.rs
+++ b/packages/rust-bindings/src/external_functions.rs
@@ -26,7 +26,8 @@ pub fn get_credential_deployment_details_ext(
     unsigned_info: &str,
     expiry: u64,
 ) -> JsResult {
-    let signatures_vec: Vec<String> = signatures.into_serde().unwrap();
+    let signatures_vec: Vec<HexString> =
+        serde_wasm_bindgen::from_value(signatures.clone()).unwrap();
     get_credential_deployment_details_aux(signatures_vec, unsigned_info, expiry).map_err(|e| {
         JsError::new(&format!(
             "Unable to get credential deployment details due to: {}",
@@ -37,7 +38,8 @@ pub fn get_credential_deployment_details_ext(
 
 #[wasm_bindgen(js_name = getDeploymentInfo)]
 pub fn get_credential_deployment_info_ext(signatures: &JsValue, unsigned_info: &str) -> JsResult {
-    let signatures_vec: Vec<String> = signatures.into_serde().unwrap();
+    let signatures_vec: Vec<HexString> =
+        serde_wasm_bindgen::from_value(signatures.clone()).unwrap();
     get_credential_deployment_info_aux(signatures_vec, unsigned_info)
         .map_err(|e| JsError::new(&format!("Unable to get credential due to: {}", e)))
 }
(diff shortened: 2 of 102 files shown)


packages/rust-bindings/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Breaking changes
+
+- Errors returned from entrypoints in `external_functions.rs` are now proper JS errors.
```
shjortConcordium, line 7: I would say that this is not a breaking change :upside_down_face:

packages/rust-bindings/src/external_functions.rs
```diff
@@ -1,12 +1,20 @@
+use std::fmt::Display;
+
 use crate::{aux_functions::*, types::*};
 use wasm_bindgen::prelude::*;
 
+type JsResult<T = JsonString> = Result<T, JsError>;
+
+fn to_js_error(error: impl Display) -> JsError { JsError::new(&format!("{}", error)) }
```
shjortConcordium, line 8: Could it make sense to just use this for everything? :thinking:

shjortConcordium (commented):
Looks good, would be nice with a test showing the error type is correct, maybe :thinking:

packages/rust-bindings/CHANGELOG.md
soerenbf: Well it is though as the type changes from `string` to `Error`? If all you do is print the error, all that will happen is that the message has "Error: " prepended, however we cannot really be sure. Isn't it best to just be honest about this?

soerenbf:
> Looks good, would be nice with a test showing the error type is correct, maybe 🤔

These tests are in place in `concordium-contracts-common` instead (see [here](https://github.com/Concordium/concordium-contracts-common/pull/87/files#diff-80b478a8fb1bd6dc4be055ecc124db18c4418c7bcbe0cd7a4a12b07934be05f4R1240-R1337))

soerenbf:
@shjortConcordium , @abizjak made a point that the release of this needs to cascade to the browser wallet as well. Could you make sure that an update to the browser wallet to include this is prioritized as well?

packages/common/test/serialization.test.ts
```diff
@@ -171,5 +171,7 @@ test('serialize type value and serializeUpdateContractParameters give same resul
 });
 
 test('serializeTypeValue throws an error if unable to serialize', () => {
-    expect(() => serializeTypeValue('test', Buffer.alloc(0))).toThrow();
+    expect(() => serializeTypeValue('test', Buffer.alloc(0))).toThrowError(
+        Error
+    );
```
shjortConcordium, line 176: :upside_down_face:

shjortConcordium (approved):

PR #200, opened 2023-06-01
Web3Id proofs

shjortConcordium:
## Purpose

Add support for web3Id proofs.

## Changes

- Added types/classes for building the new request format.
- Added `getVerifiablePresentation`.

## Checklist

      hard-to-understand areas.


packages/rust-bindings/src/aux_functions.rs
```diff
@@ -821,6 +822,76 @@ pub fn create_id_proof_aux(input: IdProofInput) -> Result<JsonString> {
...
+                    signer,
+                    values,
+                    randomness: randomness.clone(),
+                }
+            }
+        });
+    }
+    let presentation = input.request.prove(&input.global_context, commitment_inputs.into_iter());
+    Ok(json!(presentation.unwrap()).to_string())
+}
```
shjortConcordium, line 893: @abizjak This is how I "had" to do it. :thinking:

abizjak: Shouldn't we just add serde (de)serialization to commitmentinputs directly?

Why is it necessary to have this extra type?

packages/rust-bindings/src/aux_functions.rs
```diff
@@ -821,6 +822,76 @@ pub fn create_id_proof_aux(input: IdProofInput) -> Result<JsonString> {
...
+#[serde_with::serde_as]
+#[derive(SerdeDeserialize)]
+#[serde(rename_all = "camelCase", tag = "type")]
+pub enum DeserializableCommitmentInputs {
+    #[serde(rename_all = "camelCase")]
+    Account {
+        /// Issuance date of the credential that the proof is about.
+        /// This is an unfortunate name to conform to the standard, but the
+        /// meaning here really is `validFrom` for the credential.
+        issuance_date: String,
```
abizjak, line 842: We should be able to just use `chrono::DateTime<chrono::Utc>` here if we enable the `serde` feature of `chrono`.

shjortConcordium: Yes, I missed that :thinking:

packages/rust-bindings/src/aux_functions.rs
shjortConcordium: I guess we can, we just need to change CommitmentInputs to not contain references to the fields. I can probably make a PR for that.

abizjak: Right, but I like the references there. Especially for the secret key so as to avoid copying it.

That was the rationale for using them.

packages/common/src/web3Proofs.ts
```diff
@@ -0,0 +1,596 @@
...
+    schema?: VerifiableCredentialSubject
+) {
+    if (statement.type === undefined) {
+        throw new Error('Statements must contain a type field');
+    }
+    if (statement.attributeTag === undefined) {
+        throw new Error('Statements must contain an attributeTag field');
+    }
+    if (
+        existingStatements.some(
```
orhoj, line 127: Semantically I find it a little weird that `verifyAtomicStatement` does more than just that. I think the verification of no attributeTag collissions should happen elsewhere. But if it is an annoying change, then wait with this for some time in the future.

packages/common/src/web3Proofs.ts
```diff
@@ -0,0 +1,596 @@
...
+            (v) => v.attributeTag === statement.attributeTag
+        )
+    ) {
+        throw new Error('Only 1 statement is allowed for each attribute');
+    }
+
+    if (statement.attributeTag === 'id') {
```
orhoj, line 134: The `id` field was moved outside of the `attributes` portion, so I believe we allow for `id` now.

packages/common/src/web3Proofs.ts
```diff
@@ -0,0 +1,596 @@
...
+    }
+
+    /**
+     * Add to the statement, that the given attribute should be in the given range, i.e. that lower <= attribute < upper.
+     * @param attribute the attribute that should be checked
+     * @param lower: the lower end of the range, inclusive.
+     * @param upper: the upper end of the range, exclusive.
+     * @returns the updated builder
+     */
+    addRange(attribute: string, lower: string, upper: string): this {
```
orhoj, line 238: I should be able to provide integer values as input, right?

packages/common/src/web3Proofs.ts
```diff
@@ -0,0 +1,596 @@
...
+            attributeTag: attribute,
+            set,
+        };
+        this.check(statement);
+        this.statements.push(statement);
+        return this;
+    }
+
+    /**
+     * Add to the statement, that the given attribute should be one of the values in the given set.
```
orhoj, line 268: The docs here is the same as for `addMembership`. It should be the opposite of it.

packages/common/src/web3Proofs.ts
```diff
@@ -0,0 +1,596 @@
...
+function getWeb3IdCredentialQualifier(
+    validContractAddresses: ContractAddress[]
+): VerifiableCredentialQualifier {
+    return {
+        type: 'sci',
+        issuers: validContractAddresses,
+    };
+}
+
+function getAccountCredentilQualifier(
```
orhoj, line 196:
```suggestion
function getAccountCredentialQualifier(
```

orhoj (changes requested):

packages/common/src/web3Proofs.ts
```diff
@@ -0,0 +1,596 @@
...
+            );
+        }
+        if (typeof statement.upper != 'bigint') {
+            throw new Error(
+                properties.title +
+                    ' is a integer property and therefore the upper end of a range statement must be a bigint'
+            );
+        }
+    }
+    if (statement.upper < statement.lower) {
```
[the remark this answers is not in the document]
shjortConcordium: I've changed it to only check this for integer properties, and added a TODO for strings, for now :thinking:

packages/common/src/web3Proofs.ts
shjortConcordium: I added a \_not\_ :thinking:

packages/rust-bindings/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.2.0
+
+### Added
+
+- `create_web3_id_proof_ext` function.
```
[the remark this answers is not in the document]
shjortConcordium: Updated readme. We removed an entrypoint so should this be a major bump? :thinking: (Even though the function has no use)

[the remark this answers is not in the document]
orhoj: Was that version ever released to the public? Now the README also says that `verify_web3_id_credential_signature_ext` was added but also changed which is a little weird.

orhoj (approved):

packages/rust-bindings/CHANGELOG.md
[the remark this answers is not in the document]
shjortConcordium: It was weird because it was wrong, it should have said `get_verifiable_credential_signing_key_aux`.

Issue #202, opened 2023-06-07
Parse CIS2-TokenAddresses


rasmus-kirk:
We need to be able to parse CIS2 token addresses, this also includes updating the ULEB128 parts of our code.

PR #203, opened 2023-06-08
Parsing of CIS2 token addresses and improved ULEB128 functions

rasmus-kirk:
## Purpose

Addresses #202 

## Changes

Added parsing of CIS2 token addresses.

I also had to change the existing ULEB128 implementation as it was wrong, so now it correctly throws an error if the input buffer does not contain only a ULEB128 encoded number.

## Checklist

      hard-to-understand areas.


packages/common/src/uleb128.ts
```diff
@@ -1,19 +1,56 @@
...
+/**
+ * Decodes an unsigned leb128 encoded value to bigint and returns it along
+ * with the index of the end of the encoded uleb128 number + 1.
+ *
+ * @param {Buffer} bytes - The buffer to decode
+ * @param {number} index - The index where to decode at, defaults to 0
+ *
+ * @returns {bigint} the decoded bigint value.
+ */
+export function uleb128DecodeWithIndex(
```
soerenbf, line 30: We're doing a new major version soon due to other breaking changes, so I think it would make sense to consolidate these (only have `uleb128DecodeWithIndex` but rename it to `uleb128Decode`) and schedule it for release with the next breaking change.

packages/common/test/uleb128.test.ts
```diff
@@ -1,5 +1,5 @@
 import { Buffer } from 'buffer/';
-import { uleb128Decode, uleb128Encode } from '../src/uleb128';
+import { decodeLEB128, uleb128Decode, uleb128Encode } from '../src/uleb128';
```
soerenbf, line 2: what is `decodeLEB128`? forgot to fix the name after renaming? 😄

packages/common/src/cis2/util.ts
```diff
@@ -431,6 +449,40 @@ function serializeCIS2OperatorOfQuery(query: CIS2.OperatorOfQuery): Buffer {
     return Buffer.concat([owner, address]);
 }
 
+export function tokenAddressFromBase58(str: Base58String): CIS2.TokenAddress {
```
soerenbf, line 452: Remember to export this in `src/cis2/index.ts`

packages/common/CHANGELOG.md
```diff
@@ -4,9 +4,14 @@
...
 - Cost calculation for `deployModule` transaction.
 
+### Changes
+- Function `uleb128Decode` now correctly throws an error if the given input contains more than a single ULEB128 encoded number
+
 ### Added
 
 - A `parseWallet` function to parse wallet export files
+- Function `uleb128DecodeWithIndex` that can also parse more than a single ULEB128 bigint
+- Added `parseTokenAddress` to CIS2
```
soerenbf, line 14: Adapt this to reflect breaking changes.

packages/common/src/uleb128.ts
```diff
@@ -1,19 +1,57 @@
...
+ * Decodes an unsigned leb128 encoded value to bigint and returns it along
+ * with the index of the end of the encoded uleb128 number + 1.
+ *
+ * @param {Buffer} bytes - The buffer to decode
+ * @param {number} index - The index where to decode at, defaults to 0
+ *
+ * @returns {[bigint, number]} the decoded bigint value and the index of
+ * the end of the encoded uleb128 number + 1.
+ */
+export function uleb128DecodeWithIndex(
```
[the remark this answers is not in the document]
limemloh: If you keep the index argument, make sure to state that it must be non-negative, since it is not uncommon to use negative numbers to offset from the end.

packages/common/src/cis2/index.ts
```diff
@@ -1,2 +1,2 @@
-export type { CIS2 } from './util';
+export type { CIS2, tokenAddressFromBase58 } from './util';
```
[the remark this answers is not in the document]
rasmus-kirk: You're right, it was fixed

packages/common/CHANGELOG.md
```diff
@@ -2,9 +2,14 @@
 
 ## Unreleased
 
+### Changes
+- Function `uleb128Decode` now correctly throws an error if the given input contains more than a single ULEB128 encoded number
```
abizjak, line 6: Isn't it more that it actually parses it correctly, as a ULEB value should be parsed?

packages/common/src/cis2/util.ts
```diff
@@ -30,6 +35,19 @@ export namespace CIS2 {
      */
     export type Address = ContractAddress | Base58String;
 
+    /**
+     * A Token ID
+     */
+    export type TokenId = HexString;
+
+    /**
+     * A Token Address
```
abizjak, line 44: Expand a bit on what this is. Rephrasing the type name in a comment isn't very useful.

packages/common/src/cis2/util.ts
```diff
@@ -431,6 +449,38 @@ function serializeCIS2OperatorOfQuery(query: CIS2.OperatorOfQuery): Buffer {
     return Buffer.concat([owner, address]);
 }
 
+export function tokenAddressFromBase58(str: Base58String): CIS2.TokenAddress {
+    const bytes = new Buffer(bs58check.decode(str));
+
+    const firstBit = bytes[0];
```
abizjak, line 455: This is not exactly first "bit".

packages/common/src/cis2/util.ts
```diff
@@ -431,6 +449,38 @@ function serializeCIS2OperatorOfQuery(query: CIS2.OperatorOfQuery): Buffer {
     return Buffer.concat([owner, address]);
 }
 
+export function tokenAddressFromBase58(str: Base58String): CIS2.TokenAddress {
```
abizjak, line 452: I also expect a function TokenAddress -> String.

Maybe I missed it somewhere else?

rasmus-kirk: I didn't catch that you wanted both, but I'll add it

packages/common/src/uleb128.ts
```diff
@@ -1,19 +1,57 @@
...
+ * Decodes an unsigned leb128 encoded value to bigint and returns it along
+ * with the index of the end of the encoded uleb128 number + 1.
+ *
+ * @param {Buffer} bytes - The buffer to decode
+ * @param {number} index - A non-negative index to decode at, defaults to 0
+ *
+ * @returns {[bigint, number]} the decoded bigint value and the index of
+ * the end of the encoded uleb128 number + 1.
+ */
+export function uleb128DecodeWithIndex(
```
limemloh, line 32: I think this should be "defensive" and fail if the buffer is empty. Currently, it will return a decode value of 0.

limemloh (approved):
LGTM

packages/common/src/uleb128.ts
```diff
@@ -1,19 +1,57 @@
...
+
+        const c = BigInt(byte & 0x7f) << BigInt(7 * (i - index));
+        acc += c;
+
+        if ((byte & 0x80) === 0x00) {
+            break;
+        }
+    }
+
+    return [acc, nextIndex];
```
abizjak, line 52: In relation to @limemloh comment above, this parsing is slightly different than what we have elsewhere.

Namely it will successfully parse even if the buffer is empty (which is what Emil commented), but it will also do so when the buffer is not empty, but just doesn't terminate properly, so for example
```
[255u8, 255u8]
```
would be parsed successfully.

This is a bit unexpected since this is not a valid encoding of a ULEB128 value.

packages/common/src/cis2/util.ts
```diff
@@ -431,6 +450,70 @@ function serializeCIS2OperatorOfQuery(query: CIS2.OperatorOfQuery): Buffer {
...
+ *
+ * @param str A Base58 encoded token address
+ * @returns A parsed token address
+ */
+export function tokenAddressFromBase58(str: Base58String): CIS2.TokenAddress {
+    const bytes = new Buffer(bs58check.decode(str));
+
+    const firstByte = bytes[0];
+    const [index, i] = uleb128DecodeWithIndex(bytes, 1);
+    const [subindex, j] = uleb128DecodeWithIndex(bytes, i);
```
soerenbf, line 465: I would not expect a contract address to be leb128 encoded? Where does this come from?

Normally, a contract address is encoded as 8 bytes for index and subindex respectively, meaning 16 bytes for the entire contract address (not including the token ID of course). Check out the `serializeContractAddress` to see how it is serialized, and deserialize accordingly 🙂

packages/common/src/cis2/util.ts
```diff
@@ -431,6 +450,70 @@ function serializeCIS2OperatorOfQuery(query: CIS2.OperatorOfQuery): Buffer {
...
+ * @param tokenAddress A token address to convert into the base58-string format
+ * @returns The base58-formatted string
+ */
+export function tokenAddressToBase58(
+    tokenAddress: CIS2.TokenAddress
+): Base58String {
+    const firstByte = Buffer.from('02', 'hex');
+    const indexBytes = uleb128Encode(tokenAddress.contract.index);
+    const subindexBytes = uleb128Encode(tokenAddress.contract.subindex);
+    const tokenBytes = Buffer.from(tokenAddress.id, 'hex');
```
soerenbf, line 499: Also here, I would not expect a contract address should be leb128 encoded?

```suggestion
    const address = serializeContractAddress(tokenAddress.contract);
    const tokenBytes = serializeCIS2TokenId(tokenAddress.id);
```

packages/common/src/cis2/util.ts
```diff
@@ -431,6 +450,70 @@ function serializeCIS2OperatorOfQuery(query: CIS2.OperatorOfQuery): Buffer {
...
+/**
+ * Returns the Base58-encoding of the given CIS2 Token Address.
+ *
+ * @param tokenAddress A token address to convert into the base58-string format
+ * @returns The base58-formatted string
+ */
+export function tokenAddressToBase58(
+    tokenAddress: CIS2.TokenAddress
+): Base58String {
+    const firstByte = Buffer.from('02', 'hex');
```
soerenbf, line 496: What's the `02` first byte for?

packages/common/src/cis2/util.ts
```diff
@@ -431,6 +450,70 @@ function serializeCIS2OperatorOfQuery(query: CIS2.OperatorOfQuery): Buffer {
...
+        firstByte,
+        indexBytes,
+        subindexBytes,
+        tokenBytes,
+    ]);
+
+    return bs58check.encode(bytes);
+}
+
+function tokenIdFromBuffer(buffer: Buffer): CIS2.TokenId {
```
soerenbf, line 510: Could we name this `deserializeCIS2TokenId`, as there is a corresponding `serializeCIS2TokenId`?

packages/common/src/cis2/util.ts
```diff
@@ -431,6 +450,70 @@ function serializeCIS2OperatorOfQuery(query: CIS2.OperatorOfQuery): Buffer {
...
+    if (buffer.length > 255) {
```
soerenbf, line 511: Should this not be an inclusive bound?

```suggestion
    if (buffer.length > TOKEN_ID_MAX_LENGTH) {
```

packages/common/src/uleb128.ts
soerenbf: @rasmus-kirk could we add unit tests for both of these?

packages/common/src/cis2/util.ts
rasmus-kirk: It's stated that it's a version byte in the spec.

soerenbf: Alright, thanks 😄

packages/common/src/cis2/util.ts
rasmus-kirk: It comes from the spec, and it must adhere to it, right?

See:
https://proposals.concordium.software/CIS/cis-2.html#token-address

packages/common/src/cis2/util.ts
rasmus-kirk: I mean, that would make sense to me as well, but the spec disagrees.

packages/common/src/cis2/util.ts
soerenbf: Yes, sorry 🙈

packages/common/src/cis2/util.ts
abizjak: The reason is that we wanted as short as possible token addresses, which using LEB128 achieves.

And normal address serialization is using 16 bytes, and that is more convenient when handling it in the node, since parsing that is a lot simpler and predictable, and thus was used early on.

soerenbf: Thanks for clarifying!

packages/common/src/cis2/util.ts
```diff
@@ -24,7 +24,7 @@ import {
     uleb128Encode,
 } from '../uleb128';
 
-const TOKEN_ID_MAX_LENGTH = 256;
+const TOKEN_ID_MAX_LENGTH = 255;
```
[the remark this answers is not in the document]
rasmus-kirk: I set it as 255 because that was how it's done in Rust, but reading the spec I am not so sure.

@abizjak do you have insight on this?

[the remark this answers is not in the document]
soerenbf: You're right @rasmus-kirk , I had the conditionals reversed in my head.. sorry for the confusion. The spec says states the wrong thing though it seems.

packages/common/src/cis2/util.ts
```diff
@@ -431,6 +459,63 @@ function serializeCIS2OperatorOfQuery(query: CIS2.OperatorOfQuery): Buffer {
...
+ * @param tokenAddress A token address to convert into the base58-string format
+ * @returns The base58-formatted string
+ */
+export function tokenAddressToBase58(
+    tokenAddress: CIS2.TokenAddress
+): Base58String {
+    const firstByte = Buffer.from('02', 'hex');
+    const indexBytes = uleb128Encode(tokenAddress.contract.index);
+    const subindexBytes = uleb128Encode(tokenAddress.contract.subindex);
+    const tokenBytes = Buffer.from(tokenAddress.id, 'hex');
```
soerenbf, line 508: You need to prefix the serialized token ID with the length of the ID byte array. If you use this function, you will get the correct result:
 ```suggestion
    const tokenBytes = serializeCIS2TokenId(tokenAddress.id);
```

File: packages/common/src/cis2/util.ts [hunk 20]
```diff
@@ -201,6 +220,15 @@ function serializeCIS2TokenId(tokenId: HexString): Buffer {
     return packBufferWithWord8Length(serialized);
 }
 
+function deserializeCIS2TokenId(buffer: Buffer): CIS2.TokenId {
+    if (buffer.length > TOKEN_ID_MAX_LENGTH) {
+        throw Error(
+            `Token ID exceeds maximum size of ${TOKEN_ID_MAX_LENGTH} bytes`
+        );
+    }
+    return buffer.toString('hex');
```
soerenbf, line 229: The first byte of the buffer will be the length of the ID, so you need to first read the first u8 and then read the rest. Also, it would make sense to either throw an error if the length of the remaining buffer exceeds the length read from the first byte (or return the end index as done with `uleb128DecodeWithIndex`?)

soerenbf (dismissed):

packages/common/src/cis2/util.ts
soerenbf: My mistake, the spec says only the bytes of the token ID are included (which I misread to be including the length).. I think it looks all good now, sorry for the confusion.

soerenbf (approved):

abizjak (approved):

packages/common/src/uleb128.ts
```diff
@@ -1,19 +1,65 @@
...
+ * @param {number} index - A non-negative index to decode at, defaults to 0
+ *
+ * @returns {[bigint, number]} the decoded bigint value and the index of
+ * the end of the encoded uleb128 number + 1.
+ */
+export function uleb128DecodeWithIndex(
+    bytes: Buffer,
+    index = 0
+): [bigint, number] {
+    if (bytes.length === 0) {
```
limemloh, line 36: It should take the index argument into account when checking this

soerenbf: Will add a small fix for this

PR #208, opened 2023-06-21
Pv6 adaption

### LINKED ISSUE #207: Support the changes of the GRPCv2 API for the new protocol
**Task description**

The SDK should have its GRPC dependency upgraded in order to support the changes that has been made to the GRPC API for the new consensus protocol.

An overview of the changes can be found here:
https://github.com/Concordium/concordium-node/issues/870
https://github.com/Concordium/concordium-node/issues/871

soerenbf:
## Purpose

With the upcoming protocol version increment, the GRPC API changes. The changes included in this PR should align the types exposed with the new API.

The changes are implemented based on the diff seen here:

https://github.com/Concordium/concordium-grpc-api/compare/27a17bbb2c05c02ee8b90628b489962d077bdb64...13e23942735a427b188b6a27909044ff6493f50a

Closes #207 

## Checklist

      hard-to-understand areas.


packages/common/src/types.ts
```diff
@@ -284,33 +284,45 @@ export type MintDistributionV1 = MintDistributionCommon;
...
  * Used from protocol version 1-3
  */
 export interface RewardParametersV0 extends RewardParametersCommon {
     mintDistribution: MintDistributionV0;
+    gasRewards: GasRewardsV0;
 }
 
 /**
- * Used from protocol version 4
+ * Used from protocol version 4-5
```
abizjak, line 312:
```suggestion
 * Used protocol versions 4 and 5
```

packages/common/src/types.ts
```diff
@@ -345,37 +357,71 @@ export interface TimeParametersV1 {
...
+    timeoutParameters: TimeoutParameters;
+    minBlockTime: Duration;
+    blockEnergyLimit: bigint;
+}
+
+export interface FinalizationCommitteeParameters {
+    minimumFinalizers: number;
+    maximumFinalizers: number;
+    finalizerRelativeStakeThreshold: number;
+}
```
abizjak, line 379: How about some doc comments?

packages/common/src/types.ts
```diff
@@ -345,37 +357,71 @@ export interface TimeParametersV1 {
...
+export interface ConsensusParameters {
+    timeoutParameters: TimeoutParameters;
+    minBlockTime: Duration;
+    blockEnergyLimit: bigint;
+}
+
+export interface FinalizationCommitteeParameters {
+    minimumFinalizers: number;
+    maximumFinalizers: number;
+    finalizerRelativeStakeThreshold: number;
```
abizjak, line 378: It's a bit unclear what the "number" represents here. In protobuf schema you get an integer, representing parts per hundred thousand. Is that this integer, or the number `x / 100000`?

packages/common/src/types.ts
```diff
@@ -671,6 +726,30 @@ export interface ConsensusStatus {
...
 }
 
+/**
+ * Used for protocol version 1-5
+ */
+export interface ConsensusStatusV0 extends ConsensusStatusCommon {
+    /**
+     * In milliseconds
+     */
+    slotDuration: bigint;
```
abizjak, line 736: Don't you have a Duration type for this?

packages/common/src/types.ts
```diff
@@ -671,6 +726,30 @@ export interface ConsensusStatus {
...
+}
+
+export interface ConcordiumBFTStatus {
+    currentTimeoutDuration: bigint;
```
abizjak, line 740: And this.

packages/common/src/types.ts
```diff
@@ -671,6 +726,30 @@ export interface ConsensusStatus {
...
+    currentRound: bigint;
+    currentEpoch: bigint;
```
abizjak, line 742: And Epoch?

packages/common/src/types/chainUpdate.ts
```diff
@@ -87,6 +96,26 @@ export interface BakerStakeThresholdUpdate {
...
+
+export interface BlockEnergyLimitUpdate {
+    updateType: UpdateType.BlockEnergyLimit;
+    update: bigint;
+}
+
+export interface FinalizationCommitteeParametersUpdate {
+    updateType: UpdateType.FinalizationCommitteeParameters;
+    update: FinalizationCommitteeParameters;
+}
```
abizjak, line 117: Please, please add documentation for all public stuff.

packages/common/src/types.ts
soerenbf: It is `x / 100000`, which is also the case (and has been since grpcv2 was released) everywhere else the `AmountFraction` type from the GRPC API types is used. Will add a comment when I get to it.

packages/common/src/types/chainUpdate.ts
soerenbf: Yes, I agree.. I will add doc comments for all the stuff pertaining to the changes implemented here, and will add an issue to document the rest, as it will probably take a substantial amount of time.

packages/common/src/types.ts
```diff
@@ -621,20 +766,16 @@ export type BlocksAtHeightRequest =
     | AbsoluteBlocksAtHeightRequest
     | RelativeBlocksAtHeightRequest;
 
-export interface ConsensusStatus {
-    bestBlock: string;
-    genesisBlock: string;
-    currentEraGenesisBlock: string;
-    lastFinalizedBlock: string;
+export interface ConsensusStatusCommon {
```
limemloh, line 787: Some documentation would be nice here

packages/common/src/types.ts
```diff
@@ -671,6 +812,33 @@ export interface ConsensusStatus {
...
+     */
+    slotDuration: Duration;
+}
+
+export interface ConcordiumBFTStatus {
+    /**
+     * In milliseconds
+     */
+    currentTimeoutDuration: Duration;
+    currentRound: Round;
```
limemloh, line 848: Documentation plz :smile:

packages/common/src/types/chainUpdate.ts
soerenbf: Added this issue for reference: https://github.com/Concordium/concordium-node-sdk-js/issues/210

packages/common/src/versionedTypeHelpers.ts
```diff
@@ -0,0 +1,104 @@
...
+    StakePendingChange,
+    StakePendingChangeV0,
+    StakePendingChangeV1,
+    UpdateQueues,
+    UpdateQueuesV0,
+    UpdateQueuesV1,
+} from './types';
+
+export const isBakerAccountV0 = (ai: AccountInfo): ai is AccountInfoBakerV0 =>
+    (ai as AccountInfoBakerV1).accountBaker?.bakerPoolInfo === undefined;
```
limemloh, line 32: It seems wrong to cast in all of these predicates, since the point of the function is to determine the type of the argument. It will probably be fine, but you do lose some type safety here and the reader now have to think about it in order to believe it works. I generally prefer not having to think :smiling_face_with_tear:

packages/common/src/types.ts
```diff
@@ -1156,12 +1342,19 @@ export interface BakerElectionInfo {
...
+export interface ElectionInfoCommon {
     electionNonce: HexString;
     bakerElectionInfo: BakerElectionInfo[];
 }
 
+export interface ElectionInfoV0 extends ElectionInfoCommon {
+    electionDifficulty: number;
+}
+
+export type ElectionInfoV1 = ElectionInfoCommon;
```
limemloh, line 1354: Maybe we could at least add documentation to every type introduced and modified, since we have the documentation now.

limemloh (approved):
Looks great, mainly comments about missing documentation.

packages/common/src/energyCost.ts
```diff
@@ -64,7 +64,7 @@ export function getEnergyCost(
  */
 export function getExchangeRate({
     euroPerEnergy,
-    microGTUPerEuro,
+    microCCDPerEuro: microGTUPerEuro,
```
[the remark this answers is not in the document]
soerenbf: Done!

packages/common/src/types/chainUpdate.ts
rasmus-kirk: > Yes, I agree.. I will add doc comments for all the stuff pertaining to the changes implemented here, and will add an issue to document the rest, as it will probably take a substantial amount of time.

We can probably "cheat" and reuse the already quite well documented GRPC-types for most of them right? It will still take some time, but at least we don't have to write it from scratch for a good amount of them, right?

packages/common/src/versionedTypeHelpers.ts
soerenbf: I agree.. It would be better with a clear distinction using a `type` member on all of these (which also more precisely translates the GRPC type). This is more of a leftover from the grpcv1 API. I don't know if I want to include this in the scope of this feature however.

Ideally, we'd remove the deprecated `JSONRPCClient` in which case we'd only have to deal with the types from the GRPCv2 API. I think the body of work included in what you're describing would fit nicely with that?

limemloh: Yes, let's say that

soerenbf: @limemloh for reference, I've created this issue: https://github.com/Concordium/concordium-node-sdk-js/issues/212

PR #213, opened 2023-06-23
Take index into account when checking buffer length

soerenbf:


packages/common/src/uleb128.ts
```diff
@@ -33,7 +33,7 @@ export function uleb128DecodeWithIndex(
     bytes: Buffer,
     index = 0
 ): [bigint, number] {
-    if (bytes.length === 0) {
+    if (bytes[index] === undefined || bytes.subarray(0).length === 0) {
```
limemloh, line 36:
```suggestion
    if (bytes.length === index) {
```

limemloh (approved):

PR #216, opened 2023-06-26
Ensure types remain compatible with gRPC v1 API

soerenbf:
## Purpose

Forgot to take the deprecated JSON RPC client into account and the format returned from that when refactoring the types versioned by the different protocol versions. This PR reverts some changes, and makes bases the new versions on the types returned by the v1 api rather than the v2.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/nodejs/test/client.test.ts
+++ b/packages/nodejs/test/client.test.ts
@@ -13,7 +13,7 @@ import { getNodeClient } from './testHelpers';
  * remove the v1 API entirely.
  */
 
-const address = '127.0.0.1';
+const address = 'concordiumwalletnode.com';
 const port = 10000;
 
 const client = getNodeClient(address, port);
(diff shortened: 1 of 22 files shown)


packages/nodejs/test/client.test.ts
```diff
@@ -0,0 +1,289 @@
...
+let PV6_BLOCK: string | undefined = undefined;
+
+// Stagenet blocks.
+// PV4_BLOCK = 'f97d975f0e92297c51e24c3b0d8fd39dfe8e1b148d993eba6e9389d4083f7a64';
+// PV6_BLOCK = 'a3e46b2cf01d611c7507a2c2a3597da681f7b7b503b7a1f1372e661e040790f3';
+
+// Mainnet blocks.
+CHAIN_GENESIS_BLOCK =
+    '9dd9ca4d19e9393877d2c44b70f89acbfc0883c2243e5eeaecc0d1cd0503f478';
+PV4_BLOCK = '568589c9f5b3a3989c24d4c916bc2417a64c6dff6ec987595349c551a829d332';
```
shjortConcordium, line 37: Could we clean this up?

packages/nodejs/test/client.test.ts
shjortConcordium, line 1: Normally we have tests up against testnet, but these don't work for me.

soerenbf: They don't work on testnet? Did you connect to a testnet node and specify test blocks for testnet?

shjortConcordium: Let's talk about this :upside_down_face:

shjortConcordium (approved):
Looks good, just change the test to testnet.

PR #219, opened 2023-07-04
Add function to display a type schema template

DOBEN:
## Purpose

Addresses https://github.com/Concordium/concordium-contracts-common/issues/96

`Contract-commons` had new functions added to display schema templates. Making this function/feature available for the front end as well.

## Changes

- Added rust-binding to the `to_json_template()` function that can be called on a `typeSchema`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "lint-staged": {
         "*.ts": "yarn lint-fix"
     },
+    "packageManager": "yarn@3.2.1",
     "repository": {
         "type": "git",
         "url": "https://github.com/Concordium/concordium-node-sdk-js"
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "lint-staged": {
         "*.ts": "yarn lint-fix"
     },
-    "packageManager": "yarn@berry",
     "repository": {
         "type": "git",
         "url": "https://github.com/Concordium/concordium-node-sdk-js"
--- a/packages/common/src/schemaHelpers.ts
+++ b/packages/common/src/schemaHelpers.ts
@@ -42,3 +42,15 @@ export function getUpdateContractParameterSchema(
     );
     return Buffer.from(parameterSchema, 'hex');
 }
+
+/**
+ * @param rawSchema the schema for the type
+ * @returns JSON template of the schema
+ */
+export function displayTypeSchemaTemplate(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
+    rawSchema: Buffer
+): string {
+    const value = wasm.displayTypeSchemaTemplate(rawSchema.toString('hex'));
+    return value;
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -17,7 +17,6 @@
     "lint-staged": {
         "*.ts": "yarn lint-fix"
     },
-    "packageManager": "yarn@berry",
```
shjortConcordium, line 20: > Is it strictly required to use `yarn@berry`? I used my local `yarn` version but had to remove the following line to be able to commit?
> 
> ```
> "packageManager": "yarn@berry",
> ```


Lets avoid changing the configuration :upside_down_face: 
What is the error when you try to commit without removing this line? :thinking: 
(What is the result of `yarn -v` when you run it in the repo? :thinking:)

DOBEN:
```
yarn -v
3.2.1
```

Usage Error: Invalid package manager specification in package.json; expected a semver version

$ yarn ...

shjortConcordium: Hmm, I can't reproduce it.
But it seems it is not required, and mostly a feature for corepack. Maybe if we can change the line to "yarn@3.2.1" (as that seems to be the embedded version) that would be nice. :upside_down_face: 
Otherwise if that doesn't work for you, we can remove the line. :shrug:

DOBEN:
```
"packageManager": "yarn@3.2.1",
```

This works find for me. Added it.

shjortConcordium (approved):

packages/common/src/schemaHelpers.ts
```diff
@@ -42,3 +42,15 @@ export function getUpdateContractParameterSchema(
...
+ * @param rawSchema the schema for the type
+ * @returns JSON template of the schema
+ */
+export function displayTypeSchemaTemplate(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
+    rawSchema: Buffer
+): string {
+    const value = wasm.displayTypeSchemaTemplate(rawSchema.toString('hex'));
+    return value;
+}
```
shjortConcordium, line 56:
```suggestion
export function displayTypeSchemaTemplate(rawSchema: Buffer): string {
    return wasm.displayTypeSchemaTemplate(rawSchema.toString('hex'));
}
```
This can be slimmed down though. :upside_down_face:

packages/rust-bindings/src/aux_functions.rs
```diff
@@ -774,6 +774,12 @@ fn deserialize_type_value(
     }
 }
 
+pub fn display_type_schema_template_aux(schema: HexString) -> Result<JsonString> {
```
abizjak, line 777: I can see this is the same as it is done for other functions, so it is consistent, but I think we could just return the `Value` directly since wasm-bindgen can deal with those.

abizjak (approved):

PR #221, opened 2023-07-14
Add the healthCheck endpoint to the gRPCClient

shjortConcordium:
## Purpose

Add `healthCheck`.

## Changes

Added the `healthCheck` endpoint to the grpc v2 client + an example for it.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/src/GRPCClient.ts
+++ b/packages/common/src/GRPCClient.ts
@@ -10,7 +10,7 @@ import * as v2 from '../grpc/v2/concordium/types';
 import { Base58String, HexString, isRpcError } from './types';
 import { QueriesClient } from '../grpc/v2/concordium/service.client';
 import { HealthClient } from '../grpc/v2/concordium/health.client';
-import type { RpcTransport } from '@protobuf-ts/runtime-rpc';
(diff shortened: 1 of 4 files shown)


packages/common/src/GRPCClient.ts
```diff
@@ -1413,6 +1416,22 @@ export class ConcordiumGRPCClient {
...
+     * @returns a boolean indicating whether the node is healthy or not.
+     */
+    async healthCheck(): Promise<boolean> {
+        try {
+            await this.healthClient.check({});
+            return true;
+        } catch (e) {
+            return false;
+        }
+    }
```
soerenbf, line 1434: This doesn't seem right. On the GRPC types it is documented that `a return code of "OK" is used for success`. Would it not make more sense to check for this and propagate any GRPC errors thrown as well?

shjortConcordium: Well it only returns if its "OK", otherwise it will throw. (So if its catching up it will also throw an error)
I've added the error message as part of the return value.

soerenbf: Nice 😄

soerenbf (approved):

PR #225, opened 2023-07-31
Fix GRPCClient waitForTransactionFinalization

lassemoldrup:
## Purpose

The `waitForTransactionFinalization` method on `ConcordiumGRPCClient` has a bug, where some errors such as waiting for an unknown transaction do not cause the promise to reject. This is for example a problem if a transaction is submitted to one node and `waitForTransactionFinalization` is called connected to a different node. In this case, the correct solution is to wait and retry on a GRPC `NOT_FOUND` error, but as it is now, this is not possible because the promise does not reject.

## Changes

A `try` block has been expanded to catch all errors, which will then cause the promise to reject as it should.

## Checklist

### DIFF
--- a/packages/common/src/GRPCClient.ts
+++ b/packages/common/src/GRPCClient.ts
@@ -528,17 +528,19 @@ export class ConcordiumGRPCClient {
                 }, timeoutTime);
             }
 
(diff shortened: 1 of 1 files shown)


soerenbf (approved):
Looks Good 😁

Issue #214, opened 2023-06-23
Conversion helpers between `CCD` and `microCCD`


DOBEN:
**Task description**

It would be helpful to have a function `toCCD(microCCDAmount)` and `toMicroCCD(CCDAmount)` to convert between `CCD` and `microCCD` similar to `fromWei`/`toWei` functions:
https://github.com/web3/web3.js/blob/4.x/packages/web3-utils/src/converters.ts#L467

shjortConcordium:
Note that we have the following for the browser wallet, which contains something like these features:
https://github.com/Concordium/concordium-wallet-common-helpers/blob/main/src/utils/ccd.ts
(So maybe they just need to be ported/adjusted

DOBEN:
yes, exactly I am looking for the `microCcdToCcd` function if it is exported by a package already so it is available at the front end via the webSDK.

shjortConcordium:
I just meant that, for the one fixing this issue, that they don't need to create the methods from scratch. That "package" is not one we actually release/expose.

bisgardo:
The piggybank dapp has [this](https://github.com/Concordium/concordium-dapp-piggybank/blob/main/src/amount.ts):
```
export function microCcdToCcdString(amount: bigint) {
    const int = amount / BigInt(1e6);
    const frac = amount % BigInt(1e6);
    return `${int}.${frac.toString().padStart(6, '0')}`;
}
```

PR #231, opened 2023-08-07
Conversion helpers

rasmus-kirk:
## Purpose

Addresses #214 

## Changes

- `toCcd` and `fromCcd` functions to the CcdAmount class, which lets users parse and convert to ccd amounts as strings.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 9.0.1
+## Unreleased
 
 ### Added
 
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 9.0.1
+
+### Added
+
+- `toCcd` and `fromCcd` functions to the CcdAmount class, which lets users parse and convert to ccd amounts as strings.
+
 ## 9.0.0
 
 ### Breaking changes
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -4,7 +4,15 @@
 
 ### Added
 
-- `toCcd` and `fromCcd` functions to the CcdAmount class, which lets users parse and convert to ccd amounts as strings.
+Added a functions that handle conversions between CCD and micro CCD. The CCD amounts are handled as `Big`'s:
+- `toMicroCcd` returns the amount of micro CCD as a `Big`.
+- `toCcd`: returns the amount of CCD as a `Big`.
+- `fromCcd`: constructs a `CcdAmount` from an amount of CCD passed as a `Big`
+- `ccdToMicroCcd`: Converts CCD to micro CCD, both as `Big`'s
+- `microCcdToCcd`: Converts micro CCD to CCD, both as `Big`'s
+- The `CcdAmount` class constructor now also accepts a `BigSource` letting users create them from `Big`'s and strings
+
+All function parameters now also accepts strings, these strings can use comma as a decimal seperator.
 
 ## 9.0.0
 
--- a/packages/common/src/types/ccdAmount.ts
+++ b/packages/common/src/types/ccdAmount.ts
@@ -55,7 +55,7 @@ export class CcdAmount {
                 throw Error('More than one decimal seperator found!');
             }
 
-            ccd = ccd.replace('.', ',');
+            ccd = ccd.replace(',', '.');
         }
 
         const microCcd = Big(ccd).mul(Big(1000000));
--- a/packages/common/src/types/ccdAmount.ts
+++ b/packages/common/src/types/ccdAmount.ts
@@ -1,4 +1,6 @@
-import { Big } from 'big.js';
+import { Big, BigSource } from 'big.js';
+
+const MICRO_CCD_PER_CCD = 1000000;
 
 /**
  * Representation of a CCD amount.
@@ -8,7 +10,30 @@ import { Big } from 'big.js';
 export class CcdAmount {
     microCcdAmount: bigint;
 
-    constructor(microCcdAmount: bigint) {
+    /**
+     * Constructs a CcdAmount and checks that it is valid, can accept a number, string, big or bigint.
+     * Accepts strings with decimal seperators as both commas and decimals.
+     *
+     * @param microCcdAmount Theamount of micro CCD as a number, string, big or bigint.
+     * @throws If the number of commas in the input string exceeds 1
+     * @throws If a number is passed with several decimal seperators
+     * @throws If a negative amount of micro CCD is passed
+     * @throws If the micro CCD passed is greater than what can be contained in a 64-bit integer
+     */
+    constructor(microCcdAmount: bigint | BigSource) {
+        // If the input is a "BigSource" assert that the number is whole
+        if (typeof microCcdAmount !== 'bigint') {
+            microCcdAmount = newBig(microCcdAmount);
+
+            if (!microCcdAmount.mod(Big(1)).eq(Big(0))) {
+                throw Error(
+                    'Can not create CcdAmount from a non-whole number!'
+                );
+            }
+
+            microCcdAmount = BigInt(microCcdAmount.toFixed());
+        }
+
         if (microCcdAmount < 0n) {
             throw new Error(
                 'A micro CCD amount must be a non-negative integer but was: ' +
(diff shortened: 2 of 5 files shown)


packages/common/src/types/ccdAmount.ts
```diff
@@ -22,6 +24,44 @@ export class CcdAmount {
...
+    toCcd(useCommaDecimalSeperator = false): string {
+        const ccd = Big(this.microCcdAmount.toString()).div(Big(1000000));
+
+        if (useCommaDecimalSeperator) {
+            return ccd.toString().replace('.', ',');
+        } else {
+            return ccd.toString();
+        }
+    }
+
```
DOBEN, line 42: It would be good to add helper functions that are independent to the `CcdAmount` class such as

`function ccdToMicroCcd(string){ ... return string}`

`function microCcdToCcd(string){ ... return string}`

The above helper function can create an instance of the `CcdAmount` as part of its logic so (meaning it is hidden from the developer using the function that we need to deal with the `CcdAmount` class)

packages/common/src/types/ccdAmount.ts
```diff
@@ -22,6 +24,44 @@ export class CcdAmount {
...
+            const numberOfDecimalSeperators = ccd.split(',').length - 1;
+
+            if (numberOfDecimalSeperators > 1) {
+                throw Error('More than one decimal seperator found!');
+            }
+
+            ccd = ccd.replace(',', '.');
+        }
+
+        const microCcd = Big(ccd).mul(Big(1000000));
```
DOBEN, line 61: magic number 1000000.

packages/common/src/types/ccdAmount.ts
```diff
@@ -1,3 +1,5 @@
+import { Big } from 'big.js';
```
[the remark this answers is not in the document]
rasmus-kirk: The `toFraction` depends heavily on [numberStringHelpers.ts](https://github.com/Concordium/concordium-wallet-common-helpers/blob/main/src/utils/numberStringHelpers.ts) which is ~450 lines of code and doesn't seem to be used for much else. The idea is to trade the maintenance of that code with a dependency, the Sørens _seemed_ to agree, but can elaborate on this PR. 

I don't mind moving the `toFraction` and related code instead though, if people disagree on the maintenance/dependency tradeoff

packages/common/src/types/ccdAmount.ts
```diff
@@ -22,6 +24,44 @@ export class CcdAmount {
         this.microCcdAmount = microCcdAmount;
     }
 
+    /**
+     * Returns the amount of CCD as a string.
+     *
+     * @param useCommaDecimalSeperator Whether or not to use comma as a decimal seperator, defaults to false
+     * @returns The amount of CCD as a string
+     */
+    toCcd(useCommaDecimalSeperator = false): string {
```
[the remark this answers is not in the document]
rasmus-kirk: I do think `toCCD` looks nicer, but the rest of the naming in the file uses camel-case for abbreviations, so I'd rather keep consistency

packages/common/src/types/ccdAmount.ts
```diff
@@ -22,6 +24,44 @@ export class CcdAmount {
...
+    }
+
+    /**
+     * Parses an amount of CCD formatted as a string and creates a CcdAmount from it.
+     *
+     * @param ccd The amount of CCD to formatted as a string
+     * @param useCommaDecimalSeperator Whether or not to use comma as a decimal seperator, defaults to false
+     * @returns The CcdAmount object derived from the ccd input parameter
+     */
+    static fromCcd(ccd: string, useCommaDecimalSeperator = false): CcdAmount {
```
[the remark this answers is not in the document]
rasmus-kirk: Same as my other response comment

packages/common/src/types/ccdAmount.ts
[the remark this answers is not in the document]
soerenbf: I think part of the charm here is also returning the number as a `Big`. I noticed `ethers.js` do this as well (however with `BigNumber`, and it makes it a lot easier to work with the numbers coming out.

Of course one could argue that users can just pull in their numbers library of choice, but I reckon many will just use whichever we supply.

soerenbf (commented):
It seems like there are some different directions we can take this. Here are my 2 cents:

I think we might as well embrace the fact that we're pulling in a numbers library by letting the users take advantage of it instead of hiding it.

I kind of wished we had built the initial version of `CcdAmount` around this, as it provides a lot more versatility than `BigInt` and our own number helpers in the wallets, thus making it a lot easier for dApps to do calculations on our value without needing to install anything in addition to our SDK, but here we are 😄

packages/common/src/types/ccdAmount.ts
rasmus-kirk: I think this is more elegant, also not hard to implement.

packages/common/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 9.0.1
```
soerenbf, line 3: Could we do this instead?
```suggestion
## Unreleased
```

In my opinion it makes it easier to get an overview of what is released and what is not from the changelog, and then we bump all version descriptions when we release a new version.

packages/common/src/types/ccdAmount.ts
```diff
@@ -6,7 +10,30 @@
 export class CcdAmount {
     microCcdAmount: bigint;
 
-    constructor(microCcdAmount: bigint) {
+    /**
+     * Constructs a CcdAmount and checks that it is valid, can accept a number, string, big or bigint.
+     * Accepts strings with decimal seperators as both commas and decimals.
+     *
+     * @param microCcdAmount Theamount of micro CCD as a number, string, big or bigint.
+     * @throws If the number of commas in the input string exceeds 1
```
DOBEN, line 18: Does this apply also to the `dot` as a decimal separator? Then, `@throws If the number of decimal seperators in the input string exceeds 1`

packages/common/src/types/ccdAmount.ts
```diff
@@ -22,7 +49,83 @@ export class CcdAmount {
...
+        const numberOfCommaSeperators = bigSource.split(',').length - 1;
+
+        if (numberOfCommaSeperators > 1) {
+            throw Error('More than one decimal seperator found!');
+        } else if (numberOfCommaSeperators === 1) {
+            return Big(bigSource.replace(',', '.'));
+        }
+    }
+    return Big(bigSource);
+}
```
soerenbf, line 131: Here you only validate that numbers with comma separators have 0 or 1 separator, but is there validation in place for e.g. `12.0.2` in the `Big` constructor? And if so, would it not be easier to simplify this code to
```suggestion
function newBig(bigSource: BigSource): Big {
    if (typeof bigSource === 'string') {
        return Big(bigSource.replace(',', '.'));
    }
    return Big(bigSource);
}
```

packages/common/src/types/ccdAmount.ts
```diff
@@ -22,7 +49,83 @@ export class CcdAmount {
...
+     * Returns the amount of CCD as a Big.
+     *
+     * @returns The amount of CCD as a Big
+     */
+    toCcd(): Big {
+        return this.toBig().div(Big(MICRO_CCD_PER_CCD));
+    }
+
+    /**
+     * Creates a CcdAmount from a number, string, big or bigint.
```
DOBEN, line 71:
```suggestion
     * Creates a CcdAmount from a number, string, big, or bigint.
```

packages/common/src/types/ccdAmount.ts
```diff
@@ -22,7 +49,83 @@ export class CcdAmount {
         this.microCcdAmount = microCcdAmount;
     }
 
+    /**
+     * Returns the amount of micro CCD as a Big.
+     *
+     * @returns The amount of micro CCD as a Big
+     */
+    toBig(): Big {
```
soerenbf, line 57: I think the name is a bit confusing, as `Big` could represent both CCD and microCCD amounts. I would prefer a name like
```suggestion
    toMicroCcd(): Big {
```
To align with the corresponding `toCcd` function for getting CCD value wrapped in `Big`

soerenbf (commented):
Some minor stuff 😄

packages/common/src/types/ccdAmount.ts
```diff
@@ -22,7 +49,83 @@ export class CcdAmount {
...
+        return new CcdAmount(microCcd);
+    }
+
+    /**
+     * Converts an amount of CCD to micro CCD and asserts that the amount is a valid amount of CCD.
+     *
+     * @param ccd The amount of CCD
+     * @returns The amount of micro CCD as a Big
+     * @throws If the number of commas in the input string exceeds 1
+     * @throws If a number is passed with several decimal seperators
```
DOBEN, line 95: * @throws If a negative amount of micro CCD is passed
 * @throws If the micro CCD passed is greater than what can be contained in a 64-bit integer

packages/common/src/types/ccdAmount.ts
```diff
@@ -22,7 +49,83 @@ export class CcdAmount {
...
+     */
+    static ccdToMicroCcd(ccd: BigSource | bigint): Big {
+        return CcdAmount.fromCcd(ccd).toBig();
+    }
+
+    /**
+     * Converts an amount of micro CCD to CCD and asserts that the amount is a valid amount of CCD.
+     *
+     * @param microCcd The amount of micro CCD as a number, string, big or bigint.
```
DOBEN, line 104: There is a bit of inconsistency in terms of the description. E.g. (this function and the above function)

```
 @param microCcd The amount of micro CCD as a number, string, big or bigint.
 ```
 ```
  @param ccd The amount of CCD
 ```
 
 Consider making it consistent.

File: packages/common/src/types/ccdAmount.ts [hunk 13]
```diff
@@ -6,7 +10,29 @@
...
 
-    constructor(microCcdAmount: bigint) {
+    /**
+     * Constructs a CcdAmount and checks that it is valid. It accepts a number, string, big, or bigint as parameter.
+     * It can accept a string as parameter with either a comma or a dot as the decimal separator.
+     *
+     * @param microCcdAmount The amount of micro CCD as a number, string, big, or bigint.
+     * @throws If a number is passed with several decimal seperators
+     * @throws If a negative amount of micro CCD is passed
+     * @throws If the micro CCD passed is greater than what can be contained in a 64-bit integer
```
soerenbf, line 20: We could add here that it also throws if passed a non-whole number. I think we should consider consolidating all these @throws declarations into a single one such as:

```suggestion
     * @throws If an invalid micro CCD amount is passed, i.e. any value which is not an unsigned 64-bit integer
```

soerenbf (approved):
I think this looks good 😄

DOBEN (approved):

PR #233, opened 2023-08-09
common: wasmToSchema: Base custom section resolution on module version

bisgardo:
## Purpose

Ensure that `getEmbeddedSchema` (on `ConcordiumGRPCClient`) is able to resolve legacy schemas (i.e. schemas without embedded versions).

Also make parsing of contract schema more precise based by using the module version information that is now exposed by the gRPC client. This in turn also documents the relationship between module and schema versions.

## Changes

The module version is now used to determine what custom sections to look for. Added comments explain the expected results.

The extraction code is migrated from the dapp-library [`@concordium/react-components`](https://github.com/Concordium/concordium-dapp-libraries/blob/a045f3b8bc11c7389190c789072f313686b6c639/packages/react-components/src/useModuleSchemaRpc.ts). Specifically, the contents of function `getEmbeddedModuleSchema` was replaced with the contents of `findSchema` and its helper function `findCustomSections` was added and adjusted to use the SDK types.

Finally, `getEmbeddedModuleSchema` was made non-async and `wasmToSchema` removed because it was nothing but a more limited version of `getEmbeddedModuleSchema`.

Addresses https://github.com/Concordium/concordium-dapp-libraries/issues/66.


bisgardo:
@orhoj @shjortConcordium @rasmus-kirk @abizjak Do we want this? If so I'll clean it up and add some tests.

I think it would be useful and simplify https://github.com/Concordium/concordium-dapp-libraries/pull/46/files.

soerenbf:
@bisgardo Stumbled upon this PR with no reviewers and no replies. Is this still relevant? I guess if it's a general purpose thing that you've needed to implement in dapp-libraries, it might as well exist here instead?

I'm sorry this has been dangling for so long. If you deem it relevant still, I'll be happy to take a look if you get it up to speed with the main branch.

Otherwise please close it 😄

bisgardo:
> @bisgardo Stumbled upon this PR with no reviewers and no replies. Is this still relevant? I guess if it's a general purpose thing that you've needed to implement in dapp-libraries, it might as well exist here instead?
> 
> I'm sorry this has been dangling for so long. If you deem it relevant still, I'll be happy to take a look if you get it up to speed with the main branch.
> 
> Otherwise please close it 😄

@soerenbf Thanks for poking at this. I still think it's relevant :) The reason I didn't add reviewers was that I didn't consider it ready for review and I didn't want to finish it up before it was deemed worthy of doing.

I merged main into the PR branch. A couple of things have changed in the meantime and the feature seems more widely used and there seems to be some duplication that I'll try and address as part of it.

packages/sdk/src/types/VersionedModuleSource.ts
```diff
@@ -109,33 +108,61 @@ export async function parseModuleInterface(
...
+    source,
+    version,
+}: VersionedModuleSource): RawModuleSchema | null {
+    const sections = findCustomSections(
+        new WebAssembly.Module(source),
+        version
     );
-    if (versionedSchema !== null) {
-        return { type: 'versioned', buffer: versionedSchema };
+    if (!sections) {
```
soerenbf, line 121: It would be nice to spell this out explicitly, i.e. `!== undefined ...`

packages/sdk/src/pub/types.ts
```diff
@@ -25,7 +25,7 @@ export * from '../signHelpers.js';
 export * from '../versionedTypeHelpers.js';
 export * from '../accountHelpers.js';
 
-export { isHex, streamToList, wasmToSchema, unwrap } from '../util.js';
+export { isHex, streamToList, unwrap } from '../util.js';
```
soerenbf, line 28: This is a breaking change

packages/sdk/src/grpc/GRPCClient.ts
```diff
@@ -225,19 +226,16 @@ export class ConcordiumGRPCClient {
...
         blockHash?: BlockHash.Type
-    ): Promise<Uint8Array> {
-        const versionedSource = await this.getModuleSource(
-            moduleRef,
-            blockHash
-        );
-        return wasmToSchema(versionedSource.source);
+    ): Promise<RawModuleSchema | null> {
+        const source = await this.getModuleSource(moduleRef, blockHash);
+        return getEmbeddedModuleSchema(source);
```
soerenbf, line 238: This is a breaking change

packages/sdk/CHANGELOG.md
```diff
@@ -22,6 +22,16 @@
...
+### Changed
+
+- `getEmbeddedModuleSchema` now uses the module version to determine in which custom wasm sections to look for the schema.
+  It also no longer is `async` because there isn't any need for it to be so.
+- `ConcordiumGRPCClient.getEmbeddedSchema` now delegates to `getEmbeddedModuleSchema` instead of `wasmToSchema`
+  (which was removed as it was just a less capable version of `getEmbeddedModuleSchema`).
+  This means that it returns the complete `RawModuleSchema` instead of only the schema bytes.
+  It also means that it returns `null` instead of an error when no embedded schema was found.
+  
+
```
soerenbf, line 34: These are breaking changes and should be declared as such

packages/sdk/src/types/VersionedModuleSource.ts
```diff
@@ -109,33 +108,61 @@ export async function parseModuleInterface(
...
-    moduleSource: VersionedModuleSource
-): Promise<RawModuleSchema | null> {
-    const wasmModule = await WebAssembly.compile(moduleSource.source);
-    const versionedSchema = schemaBytesFromWasmModule(
-        wasmModule,
-        'concordium-schema'
+export function getEmbeddedModuleSchema({
+    source,
+    version,
+}: VersionedModuleSource): RawModuleSchema | null {
```
soerenbf, line 116: Also breaking

soerenbf (commented):
As these are breaking changes, I don't think these should be just merged to `main`. As I see it, we have 2 options: either change these to not be breaking (if possible), or follow the process outlined below.

The way we usually handle breaking changes is:

- Make a new release branch, i.e. `release/web-sdk/8`
- Target this branch for PR's implementing breaking changes
- Merge this when we're ready.

If you need to use these changes, I would suggest publishing them to npm under either the `alpha` or `beta` tag (we usually use `alpha` for internal preliminary release versions. If you intend to publish the `dapp-libraries` based on this, I would use `beta`).

bisgardo:
> As these are breaking changes, I don't think these should be just merged to `main`. As I see it, we have 2 options: either change these to not be breaking (if possible), or follow the process outlined below.
> 
> The way we usually handle breaking changes is:
> 
>     * Make a new release branch, i.e. `release/web-sdk/8`
> 
>     * Target this branch for PR's implementing breaking changes
> 
>     * Merge this when we're ready.
> 
> 
> If you need to use these changes, I would suggest publishing them to npm under either the `alpha` or `beta` tag (we usually use `alpha` for internal preliminary release versions. If you intend to publish the `dapp-libraries` based on this, I would use `beta`).

I'm in favor of keeping it a breaking change because having `ConcordiumGRPCClient.getEmbeddedSchema` only support new kinds of schemas (and actually throw an error if it's an old one) doesn't make a whole lot of sense. I would actually go a step further and remove it entirely because it's trivially expressed in terms of `getModuleSource` and `getEmbeddedModuleSchema`.

I'll let you decide what should happen. Just let me know and I'll make the relevant changes.

Btw. I'm in no rush to use this in dapp-libs. But I'm happy to prepare a PR to be merged when the time is right.

packages/sdk/src/types/VersionedModuleSource.ts
bisgardo: done

packages/sdk/src/grpc/GRPCClient.ts
bisgardo: Already mentioned in changelog.

packages/sdk/src/types/VersionedModuleSource.ts
bisgardo: Already mentioned in changelog.

packages/sdk/CHANGELOG.md
bisgardo: done

packages/sdk/src/pub/types.ts
bisgardo: Already mentioned in changelog.

packages/sdk/src/types/VersionedModuleSource.ts
bisgardo, line 116: While we're at it we could also change it to return `undefined` instead of `null`? It seems to be an outlier in returning `null`.

soerenbf: Sure, better to be consistent.

soerenbf:
> I'm in favor of keeping it a breaking change because having `ConcordiumGRPCClient.getEmbeddedSchema` only support new kinds of schemas (and actually throw an error if it's an old one) doesn't make a whole lot of sense. I would actually go a step further and remove it entirely because it's trivially expressed in terms of `getModuleSource` and `getEmbeddedModuleSchema`.
> 
> I'll let you decide what should happen. Just let me know and I'll make the relevant changes.

I think I would prefer to just keep the method on the GRPC client.

Please create a release branch for the next major version of the SDK and set it as the target branch for this. Then we'll release the changes with the next protocol update!

soerenbf (approved):
Assuming this is _not_ merged to main but to a release branch for the next major version 😉

packages/sdk/src/types/VersionedModuleSource.ts
bisgardo: done

PR #238, opened 2023-08-24
Support time-stamp attributes

shjortConcordium:
## Purpose

Add support for time-stamp attributes

## Changes

- Allow dates to be used for the statementbuilder
- Check for time-stamp attributes in validation
- Handle dates in stringify/parsing of presentations.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/src/web3ProofTypes.ts
+++ b/packages/common/src/web3ProofTypes.ts
@@ -4,10 +4,8 @@ import {
     GenericMembershipStatement,
     GenericNonMembershipStatement,
     GenericRangeStatement,
-    AtomicProof,
 } from './commonProofTypes';
 import { ContractAddress, CryptographicParameters } from './types';
-import JSONBigInt from 'json-bigint';
 
 export type AttributeType = string | bigint | Date;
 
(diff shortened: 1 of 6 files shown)


packages/common/src/web3ProofTypes.ts
```diff
@@ -163,6 +165,39 @@ export type VerifiableCredentialProof = {
     ];
 };
 
+/* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types */
+
+/**
+ * Replacer to ensure dates are stringified to the timestamp attribute format.
+ */
+export function replaceDateWithTimeStampAttribute(
```
orhoj, line 173: These two functions should perhaps be moved out of the types file and somewhere else.

orhoj (changes requested):
It would make sense to unit test the replacer/reviver function to validate that it works.

packages/common/src/web3ProofTypes.ts
shjortConcordium: Moved them together with Verifiable Presentation.
+ made tests for them
+ fixed the bug I found due to the tests :upside_down_face:

orhoj: `new Date(Date.parse(value.timestamp));`?

orhoj (approved):

PR #239, opened 2023-08-24
New consensus API extension: PR1

rasmus-kirk:
## Purpose
First PR that aims to implement the new GRPC endpoints added in the new consensus. 

**The ConcordiumClient integration tests should be rerun before merging to main!**
 
## Changes

Implemented functions:
- `getBakerEarliestWinTime`
- `getBlockCertificates`
- `getBakersRewardPeriod`
- `getWinningBakersEpoch`
- `getFirstBlockEpoch`

Changed:
- Now supports current `commissionRates` in `getPoolInfo`
- Fixed bug in `getBlockInfo` test, would not be consistent between different nodes

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -2,13 +2,14 @@
 
 ## Unreleased
 
-
 ### Added
 
 New consensus endpoints:
 - `getBakerEarliestWinTime`
 - `getBlockCertificates`
 - `getBakersRewardPeriod`
+- `getWinningBakersEpoch`
+- `getFirstBlockEpoch`
 
 ### Breaking changes
 
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -10,9 +10,6 @@ New consensus endpoints:
 - `getBakersRewardPeriod`
 - `getWinningBakersEpoch`
 - `getFirstBlockEpoch`
-
-### Breaking changes
-
 - `commissionRates` is now added to the `getPoolInfo` under `bakerPoolStatus.PoolInfo`
 
 ## 9.1.0
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## Unreleased
+
+
+### Added
+
+New consensus endpoints:
+- `getBakerEarliestWinTime`
+- `getBlockCertificates`
+- `getBakersRewardPeriod`
+
+### Breaking changes
+
+- `commissionRates` is now added to the `getPoolInfo` under `bakerPoolStatus.PoolInfo`
+
 ## 9.1.0
 
 ### Added
--- a/packages/common/src/GRPCClient.ts
+++ b/packages/common/src/GRPCClient.ts
@@ -1414,8 +1414,6 @@ export class ConcordiumGRPCClient {
 
     /**
      * Get the projected earliest time at which a particular baker will be required to bake a block.
-     * If the current consensus version is 0, this returns the status 'Unavailable', as the endpoint
-     * is only supported by consensus version 1.
      *
      * If the baker is not a baker for the current reward period, this returns a timestamp at the
      * start of the next reward period. If the baker is a baker for the current reward period, the
@@ -1427,6 +1425,8 @@ export class ConcordiumGRPCClient {
      * epoch. This is because the seed for the leader election is updated at the epoch boundary, and
      * so the winners cannot be predicted beyond that. Note that in some circumstances the returned
      * timestamp can be in the past, especially at the end of an epoch.
+     *
+     * @throws an `UNAVAILABLE` RPC error if the current consensus version is 0, as the endpoint is only supported by consensus version 1.
      *
      * @param {v1.BakerId} baker - The baker that should be queried for.
      *
@@ -1446,8 +1446,8 @@ export class ConcordiumGRPCClient {
      * certificate (if present) and epoch finalization entry (if present).
      * Note that, if the block being pointed to is not a product of ConcordiumBFT,
      * then the response will be a grpc error (invalid argument).
-     * If the endpoint is not enabled by the node, then an 'unimplemented' error
-     * will be returned.
+     *
+     * @throws an `UNIMPLEMENTED` RPC error if the endpoint is not enabled by the node.
      *
      * @param blockHash optional block hash to get the cryptographic parameters at, otherwise retrieves from last finalized block.
      *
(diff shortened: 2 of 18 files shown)


packages/common/src/GRPCClient.ts
```diff
@@ -1412,6 +1412,127 @@ export class ConcordiumGRPCClient {
...
+     * the round before the first block in the next epoch. It also indicates if a block in each
+     * round was included in the finalized chain.
+     *
+     * The following error cases are possible:
+     * @throw a `NOT_FOUND` RPC error if the query specifies an unknown block.
+     * @throw an `UNAVAILABLE` RPC error if the query is for an epoch that is not finalized in the current genesis index, or is for a future genesis index.
+     * @throw an `INVALID_ARGUMENT` RPC error if the query is for an epoch that is not finalized for a past genesis index.
+     * @throw an `INVALID_ARGUMENT` RPC error if the query is for a genesis index at consensus version 0.
+     * @throw an `INVALID_ARGUMENT` RPC error if the input `EpochRequest` is malformed.
+     * @throw an `UNAVAILABLE` RPC error if the endpoint is disabled on the node.
```
soerenbf, line 1498:
```suggestion
     * @throws a `NOT_FOUND` RPC error if the query specifies an unknown block.
     * @throws an `UNAVAILABLE` RPC error if the query is for an epoch that is not finalized in the current genesis index, or is for a future genesis index.
     * @throws an `INVALID_ARGUMENT` RPC error if the query is for an epoch that is not finalized for a past genesis index.
     * @throws an `INVALID_ARGUMENT` RPC error if the query is for a genesis index at consensus version 0.
     * @throws an `INVALID_ARGUMENT` RPC error if the input `EpochRequest` is malformed.
     * @throws an `UNAVAILABLE` RPC error if the endpoint is disabled on the node.
```

😄

packages/common/src/GRPCClient.ts
```diff
@@ -1412,6 +1412,127 @@ export class ConcordiumGRPCClient {
...
+
+    /**
+     * Get the block hash of the first finalized block in a specified epoch.
+     *
+     * The following error cases are possible:
+     * @throw - a `NOT_FOUND` RPC error if the query specifies an unknown block.
+     * @throw - an `UNAVAILABLE` RPC error if the query is for an epoch that is not finalized in the current genesis index, or is for a future genesis index.
+     * @throw - an `INVALID_ARGUMENT` RPC error if the query is for an epoch with no finalized blocks for a past genesis index.
+     * @throw - an `INVALID_ARGUMENT` RPC error if the input `EpochRequest` is malformed.
+     * @throw - an `UNAVAILABLE` RPC error if the endpoint is disabled on the node.
```
soerenbf, line 1521:
```suggestion
     * @throws - a `NOT_FOUND` RPC error if the query specifies an unknown block.
     * @throws - an `UNAVAILABLE` RPC error if the query is for an epoch that is not finalized in the current genesis index, or is for a future genesis index.
     * @throws - an `INVALID_ARGUMENT` RPC error if the query is for an epoch with no finalized blocks for a past genesis index.
     * @throws - an `INVALID_ARGUMENT` RPC error if the input `EpochRequest` is malformed.
     * @throws - an `UNAVAILABLE` RPC error if the endpoint is disabled on the node.
```

packages/common/CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
...
+- `getBakerEarliestWinTime`
+- `getBlockCertificates`
+- `getBakersRewardPeriod`
+- `getWinningBakersEpoch`
+- `getFirstBlockEpoch`
+
+### Breaking changes
+
+- `commissionRates` is now added to the `getPoolInfo` under `bakerPoolStatus.PoolInfo`
+
```
soerenbf, line 17: Damn, I must have missed that when doing the protocol version 6 type changes.. I know this is essentially breaking, but since this is basically something that should have been there all along, should we maybe document this as a "fix" instead?

Users will likely not be constructing objects of this type; merely reading from the as they are returned from the GRPC client, so I think it's fair to do this. What do you think?

soerenbf (approved):
Great job documenting all of this (even if much of it is from the generated files)!

I have some minor things you can consider, but looking good 😄

packages/common/CHANGELOG.md
abizjak: No, this is only going to be there in 6.1 node. This field is not present in version 6 node API.

soerenbf: Alright, that makes sense. I the logic of how users will be interacting with this object still stands, especially when it's also a change done in a minor release of the node. Maybe just put it under "Added"?

abizjak: I agree about the logic.

It's kind of an odd situation. From the node's perspective this is a "conservative" extension of the API, so a minor version bump makes sense.

packages/common/src/GRPCClient.ts
```diff
@@ -1412,6 +1412,127 @@ export class ConcordiumGRPCClient {
...
+     * earliest win time is projected from the current round forward, assuming that each round after
+     * the last finalized round will take the minimum block time. (If blocks take longer, or timeouts
+     * occur, the actual time may be later, and the reported time in subsequent queries may reflect
+     * this.) At the end of an epoch (or if the baker is not projected to bake before the end of the
+     * epoch) the earliest win time for a (current) baker will be projected as the start of the next
+     * epoch. This is because the seed for the leader election is updated at the epoch boundary, and
+     * so the winners cannot be predicted beyond that. Note that in some circumstances the returned
+     * timestamp can be in the past, especially at the end of an epoch.
+     *
+     * @throws an `UNAVAILABLE` RPC error if the current consensus version is 0, as the endpoint is only supported by consensus version 1.
```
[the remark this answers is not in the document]
rasmus-kirk: I agree

packages/common/src/GRPCClient.ts
```diff
@@ -1412,6 +1412,127 @@ export class ConcordiumGRPCClient {
...
+        return winTime.value;
+    }
+
+    /**
+     * For a non-genesis block, this returns the quorum certificate, a timeout
+     * certificate (if present) and epoch finalization entry (if present).
+     * Note that, if the block being pointed to is not a product of ConcordiumBFT,
+     * then the response will be a grpc error (invalid argument).
+     *
+     * @throws an `UNIMPLEMENTED` RPC error if the endpoint is not enabled by the node.
```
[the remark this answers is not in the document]
rasmus-kirk: Yes, I'll add that

PR #242, opened 2023-08-25
Initial set up of `ccd-js-gen` package

### LINKED ISSUE #236: Set up smart contract generator package
Set up the CLI tool package for generating off-chain smart contract clients.

limemloh:
## Purpose

Introduce the `ccd-js-gen` package for generating typed JavaScript code.
Closes #236

## Changes

- Introduce the `ccd-js-gen` package capable of generating simple smart contract clients (not taking advantage of the schema yet).
- Added a `ccd-js-gen` example with a script to fetch the module source for wCCD to generate a client, and a script to use this generated client.

In the SDK:
-  Added `Module` class for functionality related to smart contract modules, such as parsing the WebAssembly and interface of the module.
-  Added Smart contract related types `ContractName`, `EntrypointName` and helper functions `isInitName`, `isReceiveName`, `getContractNameFromInit` and `getNamesFromReceive`.

## Checklist

      hard-to-understand areas.


packages/common/src/types/Module.ts
```diff
@@ -0,0 +1,119 @@
...
+    /** Calculate the module reference from the module source. The module reference is cached. */
+    public getModuleRef(): ModuleReference {
+        if (this.moduleRef !== undefined) {
+            return this.moduleRef;
+        }
+        const hash = sha256([this.moduleSource]);
+        return ModuleReference.fromBytes(hash);
+    }
+
+    /** Parse the WebAssembly module in the smart contract module. The parsed module is cached. */
```
abizjak, line 63: I don't understand what "cached" means here. Cna you elaborate?

packages/common/src/types/Module.ts
```diff
@@ -0,0 +1,119 @@
...
+        return new Module(versionedModule.version, versionedModule.source);
+    }
+
+    /** Calculate the module reference from the module source. The module reference is cached. */
+    public getModuleRef(): ModuleReference {
+        if (this.moduleRef !== undefined) {
+            return this.moduleRef;
+        }
+        const hash = sha256([this.moduleSource]);
+        return ModuleReference.fromBytes(hash);
```
abizjak, line 60: Is there a reason why we don't set it
```
this.moduleRef = hash
```

?

Isn't that the meaning of "cached"?

packages/common/src/types/Module.ts
```diff
@@ -0,0 +1,119 @@
...
+    ): Module {
+        return new Module(versionedModule.version, versionedModule.source);
+    }
+
+    /** Calculate the module reference from the module source. The module reference is cached. */
+    public getModuleRef(): ModuleReference {
+        if (this.moduleRef !== undefined) {
+            return this.moduleRef;
+        }
+        const hash = sha256([this.moduleSource]);
```
abizjak, line 59: Are you sure this is correct. In Rust we have
```
    pub fn get_module_ref(&self) -> ModuleReference {
        let mut hasher = sha2::Sha256::new();
        self.serial(&mut hasher);
        ModuleReference::from(<[u8; 32]>::from(hasher.finalize()))
    }
```

so what is hashed is
- version
- length prefix (4 bytes, big endian)
- the actual source

packages/common/src/types/Module.ts
```diff
@@ -0,0 +1,119 @@
...
+        public moduleSource: Buffer
+    ) {}
+
+    /**
+     * Construct a smart contract module object from bytes, potentially read from a file.
+     * @param bytes Bytes encoding a versioned smart contract module.
+     */
+    public static fromRawBytes(bytes: Buffer): Module {
+        const version = bytes.readInt32LE(0);
+        const moduleSource = bytes.subarray(8);
```
abizjak, line 40: I think the correct thing to do would be to 
- skip the 4 bytes for the version
- read the length
- read the data of correct length.

Rather than just ignoring the length.

packages/common/src/types/Module.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+export class Module {
+    /** Cache of the parsed WebAssembly module. */
+    private wasmModule: WebAssembly.Module | undefined;
+    /** Cache of the module reference. */
+    private moduleRef: ModuleReference | undefined;
+
+    private constructor(
+        /** The version of the smart contract module. */
+        public version: number,
```
abizjak, line 29: Shouldn't we have an enum for versions?

packages/ccd-js-gen/src/lib.ts
```diff
@@ -0,0 +1,252 @@
...
+    // TODO catch if file does not exist and produce better error.
+    const fileBytes = await fs.readFile(modulePath);
+    const outputName = path.basename(modulePath, '.wasm.v1');
+    const module = SDK.Module.fromRawBytes(Buffer.from(fileBytes));
+    return generateContractClients(module, outputName, outDirPath, options);
+}
+
+/**
+ * Generate smart contract client code for a given smart contract module.
+ * @param s Buffer with bytes for the smart contract module.
```
abizjak, line 39: Shouldn't this be `scModule`?

packages/ccd-js-gen/src/lib.ts
```diff
@@ -0,0 +1,252 @@
...
+ * @param modulePath Path to the smart contract module.
+ * @param outDirPath Path to the directory to use for the output.
+ * @param options Options for generating the clients.
+ */
+export async function generateContractClientsFromFile(
+    modulePath: string,
+    outDirPath: string,
+    options: GenerateContractClientsOptions = {}
+): Promise<void> {
+    // TODO catch if file does not exist and produce better error.
```
abizjak, line 30: This should be handled in this PR I think.

packages/ccd-js-gen/src/lib.ts
```diff
@@ -0,0 +1,252 @@
...
+import * as path from 'node:path';
+import * as tsm from 'ts-morph';
+import * as SDK from '@concordium/common-sdk';
+import { Buffer } from 'buffer/';
+
+/** Output options for the generated code */
+export type OutputOptions =
+    | 'Typescript'
+    | 'Javascript'
+    | 'TypedJavascript'
```
abizjak, line 11: Why is this a useful option? Can you document the semantics of the different options.

packages/ccd-js-gen/src/lib.ts
```diff
@@ -0,0 +1,252 @@
...
+/** Convert a string in snake_case or kebab-case into camelCase. */
+function toCamelCase(str: string): string {
+    return str
+        .split(/[-_]/g)
+        .map((word, index) => (index === 0 ? word : capitalize(word)))
+        .join('');
+}
+
+/** Convert a string in snake_case or kebab-case into PascalCase. */
+function toPascalCase(str: string): string {
```
abizjak, line 250: It'd be useful to clarify why this is needed as well.

packages/common/src/types/Module.ts
limemloh: Yes, I forgot to store it. Is fixed now

packages/common/src/types/Module.ts
limemloh: I have rephrased this

packages/common/src/types/Module.ts
limemloh: makes sense

packages/ccd-js-gen/src/lib.ts
limemloh: I have added documentation for the type

packages/ccd-js-gen/src/lib.ts
```diff
@@ -0,0 +1,269 @@
...
+                    returnType: 'Promise<SDK.HexString>',
+                })
+                .setBodyText(
+                    `return this.${genericContractId}.createAndSendUpdateTransaction(
+    '${entrypointName}',
+    SDK.encodeHexString,
+    ${transactionMetadataId},
+    ${parameterId},
+    ${signerId}
+);`
```
soerenbf, line 211: Both receive functions and view functions are used to create methods here, thus creating helpers functions for signing/sending transactions for all of them. It would make more sense, if view functions would do `invokeView`.

examples/ccd-js-gen/client-tokenMetadata.ts
```diff
@@ -0,0 +1,74 @@
...
+        );
+        throw e;
+    });
+
+    const parameter = '010000'; // First 2 bytes for number of tokens to query, 1 byte for the token ID.
+    const contract = new wCCDModule.Cis2WCCD(grpcClient, contractAddress);
+
+    contract.dryRun.tokenMetadata(parameter).then((responseHex: string) => {
+        console.log({ responseHex });
+    });
```
soerenbf, line 73: By the end of this, this would be

```suggestion
    const responseHex = await contract.tokenMetadata(parameter);
    console.log({ responseHex });
```

(Notice the removal of `dryRun`, which might have gotten lost in the `await` vs `.then` change)

soerenbf (commented):
Cool to see the output client! Here are some things to consider:

packages/ccd-js-gen/src/lib.ts
limemloh: Right, I agree. The issue is that we can't really distinguish view functions from receive functions.

soerenbf: Damn... well that sucks

examples/ccd-js-gen/client-tokenMetadata.ts
```diff
@@ -0,0 +1,74 @@
+import { credentials } from '@grpc/grpc-js';
+import * as SDK from '@concordium/node-sdk';
+import meow from 'meow';
+import { parseEndpoint } from '../shared/util';
+
+//import * as wCCDModule from './lib/wCCD';
```
abizjak, line 6: What is this?

abizjak (approved):
Looks fine to me, but I am wondering why the wCCD examples are not nested under a directory that reflects that.

Now there is just `ccd-js-gen/generate.ts`. Why is this not scoped under something like
```
ccd-js-gen/wCCD/generate.ts`
```
?

examples/ccd-js-gen/client-tokenMetadata.ts
limemloh: As you know by now, the `dryRun` is needed to avoid creating the update transaction, since we can't distinguish between view and receive functions.

examples/ccd-js-gen/client-tokenMetadata.ts
limemloh: I have added a comment explaining why I left this here

packages/ccd-js-gen/src/lib.ts
```diff
@@ -0,0 +1,269 @@
...
+                    returnType: 'Promise<SDK.HexString>',
+                })
+                .setBodyText(
+                    `return this.${genericContractId}.invokeView(
+    '${entrypointName}',
+    SDK.encodeHexString,
+    (hex: SDK.HexString) => hex,
+    ${parameterId},
+    ${blockHashId}
+);`
```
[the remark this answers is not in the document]
limemloh: As you already know by now (adding this for anyone else), we are not able to distinguish view functions from receive functions.

packages/ccd-js-gen/src/lib.ts
```diff
@@ -0,0 +1,275 @@
...
+    ${parameterId},
+    ${signerId}
+);`
+                );
+            const blockHashId = 'blockHash';
+            dryRunClassDecl
+                .addMethod({
+                    docs: [
+                        `Dry run a message to the '${entrypointName}' entrypoint of the '${contract.contractName}' contract`,
+                    ],
```
soerenbf, line 224: Should we add more comprehensive documentation to each method? Would be nice if the result would be in the form of

```javascript
/**
 * Dry run a message to the '<entrypointName>' entrypoint of the '<contract.contractName>' contract
 *
 * @param {SDK.HexString} parameter - Hex encoded parameter for entrypoint
 * @param {SDK.HexString} [blockHash] - Block hash of the block to invoke entrypoint at
 * 
 * @throws If the entrypoint is not successfully invoked.
 * 
 * @returns {SDK.HexString} Hex encoded response
 */
```

packages/ccd-js-gen/src/lib.ts
```diff
@@ -0,0 +1,275 @@
...
+
+        for (const entrypointName of contract.entrypointNames) {
+            const transactionMetadataId = 'transactionMetadata';
+            const parameterId = 'parameter';
+            const signerId = 'signer';
+            classDecl
+                .addMethod({
+                    docs: [
+                        `Send a message to the '${entrypointName}' entrypoint of the '${contract.contractName}' contract.`,
+                    ],
```
soerenbf, line 191: And correspondingly also here.

Also, what happens if you try to send a view function invocation as a contract update? Should we explain somewhere that this is a limitation of the current implementation?

soerenbf (commented):
I still think it would be nice to add the option for explicitly specifying which entrypoints (with receive/view being specified separately) as it's a 1-time setup thing.

We could even provide feedback in the terminal as to which entrypoints have been

- Included in the generated client
- Excluded from the generated client
- Not found (maybe throw an error here?).

The **advantages** here being:

- Users will not try sending view invocations as transactions by mistake (e.g. it would be reasonable to expect, as a user, that to get cis2 balance you would do `client.balanceOf` instead of `client.dryRun.balanceOf`).
- Possibility to include only the code you need in your bundle (tree shaking does not work on classes, as you know)

**Disadvantages**:
- Users will have to change generate config when contract changes, as opposed to only changing the code interfacing with the generated client
- Users will have to know up-front which entrypoints are view/receive functions instead of through trial/error when coding

What are your thoughts on this? Of course this also depends on the expected timeline for being able to identify view/receive functions in a contract

packages/ccd-js-gen/src/lib.ts
limemloh: Just to answer your question: The chain and smart contract is not really aware of 'view functions'. A receive-function of a smart contract have the option to return a value, and when invoked using `invokeInstance` it will run the code locally just in the node and provide you this return value, which is then used for something like view functions. Sending a contract update transaction to the chain will just ignore the return value.
In other words, sending update transactions to view functions will be treated as any other receive function, but since a view function usually don't do any side-effects it will just cost the caller some CCD.

limemloh:
> I still think it would be nice to add the option for explicitly specifying which entrypoints (with receive/view being specified separately) as it's a 1-time setup thing.
> 
> We could even provide feedback in the terminal as to which entrypoints have been
> 
>     * Included in the generated client
> 
>     * Excluded from the generated client
> 
>     * Not found (maybe throw an error here?).
> 
> 
> The **advantages** here being:
> 
>     * Users will not try sending view invocations as transactions by mistake (e.g. it would be reasonable to expect, as a user, that to get cis2 balance you would do `client.balanceOf` instead of `client.dryRun.balanceOf`).
> 
>     * Possibility to include only the code you need in your bundle (tree shaking does not work on classes, as you know)
> 
> 
> **Disadvantages**:
> 
>     * Users will have to change generate config when contract changes, as opposed to only changing the code interfacing with the generated client
> 
>     * Users will have to know up-front which entrypoints are view/receive functions instead of through trial/error when coding
> 
> 
> What are your thoughts on this? Of course this also depends on the expected timeline for being able to identify view/receive functions in a contract

Great points you are bringing up, and I think we need to reconsider some of my initial design decisions, especially related to producing tree-shakable code. Regarding your suggestion of taking a contract description when generating the client, I would suggest keeping this for a future iteration of the tool and let it generate everything for now.

Since the primary goal of this PR is to set up the project and give a starting point, I suggest we get it merged and do the changes afterward.

packages/ccd-js-gen/src/lib.ts
```diff
@@ -0,0 +1,290 @@
...
+                },
+                {
+                    docs: ['Generic contract client used internally.'],
+                    scope: tsm.Scope.Private,
+                    name: genericContractId,
+                    type: 'SDK.Contract',
+                },
+                {
+                    docs: ['Dry run entrypoints of the smart contract.'],
+                    scope: tsm.Scope.Public,
```
soerenbf, line 135:
```suggestion
                    scope: tsm.Scope.Public,
                    isReadonly: true,
```

?

soerenbf (approved):

PR #244, opened 2023-08-25
Compare string attributes

shjortConcordium:
## Purpose

Add helpers to compare string attributes + use them for validating range statements.

## Changes

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/common/src/web3IdHelpers.ts
+++ b/packages/common/src/web3IdHelpers.ts
@@ -29,15 +29,19 @@ export function verifyWeb3IdCredentialSignature(
  * if a > b then compareStringAttributes(a,b) = 1;
  */
 export function compareStringAttributes(a: string, b: string): number {
-    return wasm.compareStringAttributes(a,b);
+    return wasm.compareStringAttributes(a, b);
 }
 
 /**
  * Given a string attribute value and a range [lower, upper[, return whether value is in the range, when converted into field elements.
  */
-export function isStringAttributeInRange(value: string, lower: string, upper: string): boolean {
+export function isStringAttributeInRange(
+    value: string,
+    lower: string,
+    upper: string
+): boolean {
     const lowCmp = compareStringAttributes(value, lower);
-    if (lowCmp < 0 ) {
+    if (lowCmp < 0) {
         return false;
     }
     const upCmp = compareStringAttributes(value, upper);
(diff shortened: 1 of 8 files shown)


packages/common/src/web3IdHelpers.ts
```diff
@@ -21,3 +21,29 @@ export function verifyWeb3IdCredentialSignature(
...
+ * if a > b then compareStringAttributes(a,b) = 1;
+ */
+export function compareStringAttributes(a: string, b: string): number {
+    return wasm.compareStringAttributes(a, b);
+}
+
+/**
+ * Given a string attribute value and a range [lower, upper[, return whether value is in the range, when converted into field elements.
+ */
+export function isStringAttributeInRange(
```
orhoj, line 38: Make unit tests for the edge cases where `value === lower` and `value === upper`.

packages/common/src/web3Proofs.ts
```diff
@@ -103,7 +109,7 @@ function verifyRangeStatement(
             }
         };
 
-        if (properties.type === 'string' && properties.format === 'date-time') {
+        if (isTimeStamp(properties)) {
```
orhoj, line 112: I am unsure if we actually agreed to parse it is a date when `format = date-time` or if we decided to have a custom `type = timestamp` for this.

orhoj (changes requested):

packages/common/src/web3Proofs.ts
shjortConcordium: Fixed this.

packages/common/src/web3IdHelpers.ts
shjortConcordium: Added these.
What would we expect if `value === lower === upper`? :thinking:

orhoj: That is invalid input as it would result in an empty interval as `[a, a) = ∅`.

orhoj (changes requested):

packages/common/src/web3ProofTypes.ts
```diff
@@ -39,15 +39,32 @@ export type Web3IdProofInput = {
...
+    description?: string;
+    type: 'string' | 'integer';
     format?: string;
 };
 
+export type PropertyDetails = SimplePropertyDetails | TimestampPropertyDetails;
+
 type IndexDetails = {
-    title: 'id';
+    title: string;
```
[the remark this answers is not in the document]
shjortConcordium: I don't think the title needs to be id, at least it is not in our example :thinking:  
https://github.com/Concordium/concordium-web3id/blob/main/examples/json-schemas/education-certificate/JsonSchema2023-education-certificate.json

orhoj (approved):

PR #252, opened 2023-08-30
Add module client to SDK and ccd-js-gen

limemloh:
## Purpose

Add type `DeployedModule` representing a generic module on chain to the SDK.
Generate a tailored `DeployedModule` client from a smart contract module source.

## Checklist

      hard-to-understand areas.


packages/common/src/GenericContract.ts
```diff
@@ -223,6 +238,32 @@ class ContractBase<E extends string = string, V extends string = string> {
...
+     * Get information on this smart contract instance.
+     */
+    public async getInstanceInfo(blockHash?: string): Promise<InstanceInfo> {
+        return this.grpcClient.getInstanceInfo(this.contractAddress, blockHash);
+    }
+
+    /**
+     * Check if the smart contract instance exists on the blockchain.
+     * Optionally a module reference can be provided to check if the contract instance uses this module.
+     * @param {ContractCheckOnChainOptions} options Options for checking information on chain.
```
soerenbf, line 251: `@throws` please 😉

packages/common/src/GenericContract.ts
```diff
@@ -223,6 +238,32 @@ class ContractBase<E extends string = string, V extends string = string> {
         return getContractName(instanceInfo);
     }
 
+    /**
+     * Get information on this smart contract instance.
```
soerenbf, line 242: `@throws` similar to `getContractName`

packages/common/src/DeployedModule.ts
```diff
@@ -0,0 +1,115 @@
...
+    /** Private constructor to enforce creating objects using a static method. */
+    private constructor(
+        /** The gRPC connection used by this object */
+        public grpcClient: ConcordiumGRPCClient,
+        /** The reference for this module */
+        public moduleReference: ModuleReference
+    ) {}
+
+    /** Create a new `GenericModule` instance for interacting with a smart contract module on chain.
+     * This function ensures the module is already deployed on chain otherwise produces an error. */
```
soerenbf, line 40: Would be nice with an explicit `@throws` declaration instead of the "otherwise produces and error".

packages/common/src/DeployedModule.ts
```diff
@@ -0,0 +1,115 @@
...
+     * The caller must ensure that the smart contract module is already deployed on chain. */
+    public static createUnchecked(
+        grpcClient: ConcordiumGRPCClient,
+        moduleReference: ModuleReference
+    ): DeployedModule {
+        return new DeployedModule(grpcClient, moduleReference);
+    }
+
+    /** Check if this module is deployed to the chain. */
+    public async checkOnChain(blockHash?: string): Promise<void> {
```
soerenbf, line 60: Would it make more sense to have the signature be

```suggestion
    public async checkOnChain(blockHash?: string): Promise<boolean> {
```

?

Maybe also add a `@throws` declaration for GRPC errors?

packages/common/src/DeployedModule.ts
```diff
@@ -0,0 +1,115 @@
...
+        await this.getModuleSource(blockHash);
+    }
+
+    /** Get the module source of the deployed smart contract module. */
```
soerenbf, line 64: `grpcClient.getModuleSource` throws, which should probably be mentioned here as well.

soerenbf (approved):
Being a little nit-picky with the `@throws` declarations. It's not that I have been very consistent with this myself, but I think it makes a lot of sense as you (as a developer) are going in completely blind (or JS-naive? 😄 ) if it isn't mentioned anywhere.

packages/common/src/DeployedModule.ts
```diff
@@ -0,0 +1,115 @@
...
+ * An update transaction without header.
+ */
+export type ContractInitTransaction = {
+    /** The type of the transaction, which will always be of type {@link AccountTransactionType.InitContract} */
+    type: AccountTransactionType.InitContract;
+    /** The payload of the transaction, which will always be of type {@link InitContractPayload} */
+    payload: InitContractPayload;
+};
+
+/** Type representing a smart contract module deployed on chain. */
```
soerenbf, line 29: Would be good to add a `@template C - The contract name` declaration here

packages/common/src/DeployedModule.ts
```diff
@@ -0,0 +1,115 @@
...
+    /** Create a new `GenericModule` instance for interacting with a smart contract module on chain.
+     * This function ensures the module is already deployed on chain otherwise produces an error. */
+    public static async create(
+        grpcClient: ConcordiumGRPCClient,
+        moduleReference: ModuleReference
+    ): Promise<DeployedModule> {
+        const mod = new DeployedModule(grpcClient, moduleReference);
+        await mod.checkOnChain();
+        return mod;
+    }
```
soerenbf, line 48:
```suggestion
    public static async create<C extends string = string>(
        grpcClient: ConcordiumGRPCClient,
        moduleReference: ModuleReference
    ): Promise<DeployedModule<C>> {
        const mod = new DeployedModule<C>(grpcClient, moduleReference);
        await mod.checkOnChain();
        return mod;
    }
```

+ `@template C` declaration

packages/common/src/DeployedModule.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    /** Create a new `GenericModule` instance for interacting with a smart contract module on chain.
+     * The caller must ensure that the smart contract module is already deployed on chain. */
+    public static createUnchecked(
+        grpcClient: ConcordiumGRPCClient,
+        moduleReference: ModuleReference
+    ): DeployedModule {
+        return new DeployedModule(grpcClient, moduleReference);
+    }
```
soerenbf, line 57:
```suggestion
    public static createUnchecked<C extends string = string>(
        grpcClient: ConcordiumGRPCClient,
        moduleReference: ModuleReference
    ): DeployedModule<C> {
        return new DeployedModule<C>(grpcClient, moduleReference);
    }
```

+ `@template C` declaration

soerenbf (commented):
Would be nice with an example on both generating and using this as well, or doesn't that make sense at this point?

It helped a lot in reviewing the code as it was easy to see the output 🙂

packages/common/src/DeployedModule.ts
limemloh: The reason for throwing instead of the boolean is to be able to communicate different reasons for failing the check. This is mainly needed in `checkOnChain` for the `GenericContract`, so this is just to follow the signature

limemloh:
Moved to be part of https://github.com/Concordium/concordium-node-sdk-js/pull/262

PR #269, opened 2023-09-20
Publish as ESM only

soerenbf:
## Purpose

This PR changes the library to be published _only_ as ESM, which is the recommended module strategy for both bundlers and node.

It doesn't change anything in terms of compatibility. The only thing this changes is, that it requires users developing for nodeJS to configure their project to use ES modules instead of commonJS. This is documented in the published documentation. For good measure, this is done by setting a `package.json` field:

```json
{
  ...
  "type": "module",
  ...
}
```

Sticking to supporting one standard might also make it easier to transition to making the `web-sdk` platform agnostic, thus dropping (or deprecating) the `node-sdk` altogether.

## Changes

- Add file endings to all relative imports and directory imports
- Output GRPC as typescript + add file endings for imports on generation

## Checklist

      hard-to-understand areas.


limemloh (approved):

packages/nodejs/src/client.ts
```diff
@@ -621,20 +595,16 @@ export default class ConcordiumNodeClient {
...
         }
-        const blockHashObject = new BlockHash();
-        blockHashObject.setBlockHash(blockHash);
-
-        const response = await this.sendRequest(
-            this.client.getInstances,
-            blockHashObject
-        );
+        const input: BlockHash = { blockHash };
+        const { value } = await this.client.getAnonymityRevokers(input)
```
limemloh, line 599: I don't think this is right

packages/rust-bindings/package.json
```diff
@@ -67,6 +67,6 @@
     },
     "dependencies": {
         "buffer": "^6.0.3",
-        "webpack": "^5.72.0"
+        "webpack": "^5.88.2"
```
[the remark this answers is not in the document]
soerenbf: Good point..

packages/nodejs/src/client.ts
soerenbf: Good catch, I'll check them all again for good measure 😄

Issue #250, opened 2023-08-29
Remove `buffer` as peer dependency


limemloh:
**Description**

When using the SDK, the user will need to install [`buffer`](https://www.npmjs.com/package/buffer) as a dependency, since types from this library is exposed in the public API of our SDK.
Currently, we don't list it as a [`peerDependency`](https://nodejs.org/en/blog/npm/peer-dependencies) in `package.json` and we could fix that, but IMO it would be more appropriate to only expose [`ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) and only use buffer internally.

soerenbf:
I agree. We can still resolve to the browser-specific implementation in the web target, but it does feel awkward to have to use "our" version of buffer even in the node SDK, when the platform has it as a native type as well (and if you use the native type with typescript, you get type errors).

PR #273, opened 2023-09-25
Remove dependency on Buffer being used externally

soerenbf:
## Purpose

Ensure `Buffer` is not required as an external dependency. Also removes all imports of `buffer/index.js` in node-sdk.

Partially implements #250.


limemloh (approved):
Great!

PR #274, opened 2023-09-26
Remove grpc v1

### LINKED ISSUE #212: Remove deprecated JSON RPC client
**Task description**
The `JsonRpcClient` has been marked deprecated since the release of the gRPC v2 API of the node. Currently, we translate all types from the gRPC v2 API to their gRPC v1 API correspondants. If we could remove the need to support the first version we could make the code base easier to maintain.

Obviously, these changes will be breaking, and should be scheduled with a major release.

**Clarifications**
- [x] Is there any reason to still keep the JSON RPC client? Do we need to keep supporting the GRPCv1 API?

**Sub-tasks**
- [x] Remove `JsonRpcClient` and types only associated with this
- [x] Change top-level types versioned by different protocol versions to explicitly state a version
  - ~~We could even try to eliminate type translations altogether, and just use the generated GRPC types. Only issue with this is that the generated types seem to be all optional, where in reality they're not. What we could do here is simply override the optional properties in question with required ones.~~
    - Not doing this, as it will decrease developer experience (types are worse) and will cause a lot of breaking changes, requiring a lot of changes in our own consumers of the SDK.
  - [x] `ChainParameters`
  - [x] `ElectionInfo`
  - [x] `BlockInfo`
  - [x] `ConsensusStatus`

**From #135**
  - [ ]  ~~Remove TransactionSummaryType~~
  - [x] Update delegatorId fields in event data to be bigints
  - [x] chainParameters.foundationAccount should not be optional

soerenbf:
## Purpose

Removes GRPC v1 clients (i.e. `ConcordiumNodeClient` from `@concordium/node-sdk` and `JsonRpcClient` from `@concordium/web-sdk`).

Closes #212

## Changes

- Remove `JsonRpcClient` and `ConcordiumNodeClient` along with functionality only used by these.
- Delete types marked as deprecated.
- Add version field to types versioned by protocol version

## Checklist

      hard-to-understand areas.


limemloh (approved):
Maybe rebase on `api-strict-types`? The unit test should not fail then

PR #276, opened 2023-10-02
Use CcdAmount, TransactionExpiry and ModuleRef in api

### LINKED ISSUE #253: Use wrapper types for ccd amounts and transaction expiries
The following will cause breaking changes, and should be included in a new major version.

In the type `ContractTransactionMetadata` in `common/src/GenericContract.ts`, it would be good usability if users could supply ccd amounts as `CcdAmount`s instead of `bigint`s. The same can be said for the expiry date, which could be a `TransactionExpiry` instead.

This would allow us to utilize guarantees provided by these types.

It would make sense to check the codebase for other places this could be applied.

limemloh:
## Purpose

Closes #253.
Use CcdAmount, TransactionExpiry and ModuleReference more consistently in API.

## Changes

- Change `CcdAmount`, `TransactionExpiry` and `ModuleReference` to follow the module pattern used by other types.
- Constructing `TransactionExpiry` no longer checks whether the date is in the past, since the check is only relevant at the point of sending the transaction to chain.
- Found and fixed a few more instances of `string` instead of `AccountAddress` in the API.

## Checklist

      hard-to-understand areas.


soerenbf (approved):
Looks good (although examples typecheck doesn't run successfully)!

PR #279, opened 2023-10-05
Types json serialize

soerenbf:
## Purpose

Adds functionality for formatting/parsing strong types introduced as JSON.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/src/types/AccountAddress.ts
+++ b/packages/sdk/src/types/AccountAddress.ts
@@ -1,22 +1,22 @@
 import bs58check from 'bs58check';
 import { Buffer } from 'buffer/index.js';
 import type * as Proto from '../grpc-api/v2/concordium/types.js';
-import { TypeBase, TypedJsonDiscriminator, fromTypedJson } from './util.js';
+import { TypeBase, TypedJsonDiscriminator, makeFromTypedJson } from './util.js';
 import { Base58String } from '../types.js';
 
 /**
  * The {@linkcode TypedJsonDiscriminator} discriminator associated with {@linkcode Type} type.
  */
 export const JSON_TYPE = TypedJsonDiscriminator.AccountAddress;
-type Json = Base58String;
+type Serializable = Base58String;
 
 /**
  * Representation of an account address, which enforces that it:
  * - Is a valid base58 string with version byte of 1.
  * - The base58 string is a length of 50 (encoding exactly 32 bytes).
  */
-class AccountAddress extends TypeBase<Json> {
-    protected jsonType = JSON_TYPE;
+class AccountAddress extends TypeBase<Serializable> {
+    protected typedJsonType = JSON_TYPE;
 
     constructor(
         /** The account address represented in base58check. */
@@ -27,7 +27,7 @@ class AccountAddress extends TypeBase<Json> {
         super();
     }
 
-    protected get jsonValue(): Json {
+    protected get serializableJsonValue(): Serializable {
         return this.address;
     }
 }
@@ -212,4 +212,4 @@ export function equals(left: AccountAddress, right: AccountAddress): boolean {
  * @throws {TypedJsonParseError} - If unexpected JSON string is passed.
  * @returns {AccountAddress} The parsed instance.
  */
-export const fromJSON = fromTypedJson(JSON_TYPE, fromBase58);
+export const fromTypedJSON = makeFromTypedJson(JSON_TYPE, fromBase58);
--- a/packages/sdk/src/types/AccountAddress.ts
+++ b/packages/sdk/src/types/AccountAddress.ts
@@ -1,7 +1,11 @@
 import bs58check from 'bs58check';
 import { Buffer } from 'buffer/index.js';
 import type * as Proto from '../grpc-api/v2/concordium/types.js';
-import { TypeBase, TypedJsonDiscriminator, makeFromTypedJson } from './util.js';
+import {
+    TypedJson,
+    TypedJsonDiscriminator,
+    makeFromTypedJson,
+} from './util.js';
 import { Base58String } from '../types.js';
 
 /**
@@ -15,21 +19,15 @@ type Serializable = Base58String;
  * - Is a valid base58 string with version byte of 1.
  * - The base58 string is a length of 50 (encoding exactly 32 bytes).
  */
-class AccountAddress extends TypeBase<Serializable> {
-    protected typedJsonType = JSON_DISCRIMINATOR;
-
+class AccountAddress {
+    /** Having a private field prevents similar structured objects to be considered the same type (similar to nominal typing). */
+    private __type = JSON_DISCRIMINATOR;
     constructor(
         /** The account address represented in base58check. */
         public readonly address: string,
         /** The account address represented in bytes. */
         public readonly decodedAddress: Uint8Array
-    ) {
-        super();
-    }
-
-    protected get serializable(): Serializable {
-        return this.address;
-    }
+    ) {}
 }
 
 /**
@@ -38,8 +36,16 @@ class AccountAddress extends TypeBase<Serializable> {
  * - The base58 string is a length of 50 (encoding exactly 32 bytes).
  */
 export type Type = AccountAddress;
-export const instanceOf = (value: unknown): value is AccountAddress =>
-    value instanceof AccountAddress;
+
+/**
+ * Type predicate for {@linkcode Type}
+ *
+ * @param value value to check.
+ * @returns whether `value` is of type {@linkcode Type}
+ */
+export function instanceOf(value: unknown): value is AccountAddress {
+    return value instanceof AccountAddress;
+}
 
 /**
  * Type guard for AccountAddress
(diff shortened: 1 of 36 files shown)


packages/sdk/src/types/util.ts
```diff
@@ -0,0 +1,163 @@
...
+    /** The type discriminator */
+    ['@type']: TypedJsonDiscriminator;
+    /** The serializable type value */
+    value: V;
+}
+
+/**
+ * Base class for concordium domain types
+ *
+ * @template D - The JSON discriminator
```
limemloh, line 42: No type parameter named `D`
```suggestion
```

packages/sdk/src/types/AccountAddress.ts
```diff
@@ -24,6 +38,8 @@ class AccountAddress {
  * - The base58 string is a length of 50 (encoding exactly 32 bytes).
  */
 export type Type = AccountAddress;
+export const instanceOf = (value: unknown): value is AccountAddress =>
```
limemloh, line 41: Documentation would be nice here. Consider defining using `function ...` to follow the style of the rest of the file

packages/sdk/src/types/util.ts
```diff
@@ -0,0 +1,163 @@
...
+ * @template V - The serializable JSON value
+ */
+export abstract class TypeBase<V> implements ToTypedJson<V> {
+    protected abstract typedJsonType: TypedJsonDiscriminator;
+    protected abstract get serializable(): V;
+
+    public toTypedJSON(): TypedJson<V> {
```
limemloh, line 49: Since we are not using the built-in `toJSON` this might as well be a separate function, so it can be tree-shaken.
It is not really any different from the other serializations we have (protobuf, schemaJson, bytes and so on)

packages/sdk/src/types/util.ts
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Base class for concordium domain types
+ *
+ * @template D - The JSON discriminator
+ * @template V - The serializable JSON value
+ */
+export abstract class TypeBase<V> implements ToTypedJson<V> {
+    protected abstract typedJsonType: TypedJsonDiscriminator;
+    protected abstract get serializable(): V;
```
limemloh, line 47: I think these abstract fields are a bit too much abstraction without any benefit. The function needs to return `TypedJson<V>` so all of this is already covered IMHO.

packages/sdk/src/types/AccountAddress.ts
```diff
@@ -24,6 +38,8 @@ class AccountAddress {
...
+    value instanceof AccountAddress;
 
 /**
  * Type guard for AccountAddress
```
limemloh, line 45: We probably need to remove this function `isAccountAddress` since it can cast plain objects to be of `AccountAddress` without actually being an object of the class.

packages/sdk/src/types/util.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+/**
+ * Creates a function to convert {@linkcode TypedJson} to their corresponding type instance.
+ *
+ * @template V - The JSON value
+ * @template T - The type returned
+ *
+ * @param {D} expectedTypeDiscriminator - The discriminator expected in the JSON string parsed
```
limemloh, line 146:
```suggestion
 * @param {TypedJsonDiscriminator} expectedTypeDiscriminator - The discriminator expected in the JSON string parsed
```

packages/sdk/src/types/util.ts
```diff
@@ -0,0 +1,163 @@
+/**
+ * Discriminator for {@linkcode TypedJson}. The member used to identify each type is
+ * exported from each type module and can be accessed through named export `JSON_DISCRIMINATOR`.
+ */
+export enum TypedJsonDiscriminator {
```
limemloh, line 5: I like this approach over the `_nominal` that I introduced, since the type error message becomes better. :+1:

packages/sdk/src/types/util.ts
```diff
@@ -0,0 +1,163 @@
...
+    TransactionExpiry = 'ccd_transaction_expiry',
+    TransactionHash = 'ccd_transaction_hash',
+}
+
+/**
+ * Type describing the JSON representation of strong types used in the SDK.
+ *
+ * @template V - The serializable JSON value
+ */
+export interface TypedJson<V> {
```
limemloh, line 32: I suggest using type aliases to avoid accidental declaration merging:
```suggestion
export type TypedJson<V> = {
```

limemloh (changes requested):

packages/sdk/src/types/util.ts
```diff
@@ -0,0 +1,163 @@
...
+ * @throws {TypedJsonParseError} {@linkcode TypedJsonParseError} if the returned function fails to parse the passed value.
+ *
+ * @returns The JSON parser function
+ */
+export function makeFromTypedJson<V, T>(
+    expectedTypeDiscriminator: TypedJsonDiscriminator,
+    toType: (value: V) => T
+) {
+    return ({ ['@type']: type, value }: TypedJson<V>): T | V => {
+        if (!type || !value) {
```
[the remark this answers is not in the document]
soerenbf: Yes I agree. Thanks for pointing this out

packages/sdk/src/types/json.ts
```diff
@@ -0,0 +1,182 @@
...
+export function jsonStringify(
+    input: any,
+    replacer?: any,
+    space?: string | number
+): string {
+    /**
+     * Recursively maps concordium domain types to values that can be revived into their original types.
+     */
+    const mapValues = (v: any): any => {
+        if (v == undefined || typeof v !== 'object') return v;
```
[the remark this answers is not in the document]
soerenbf: Well that's a matter of opinion I think. This is the only case where I think loose equality is acceptable (I usually do `v == null` though as I think that makes more sense in terms of understanding the boundaries of the loose equality check), and my impression is that it's a common use case for loose equality.

I don't mind the explicit version either, so will change it to that.

packages/sdk/src/types/json.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    return undefined;
+}
+
+/**
+ * Acts as an inverse for {@linkcode jsonStringify}
+ */
+export function jsonParse(
+    input: string | undefined,
```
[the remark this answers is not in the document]
soerenbf: This is in line with the `JSON.parse` does as well, so I agree that this makes more sense.

packages/sdk/src/types/AccountAddress.ts
```diff
@@ -190,3 +206,12 @@ export function toProto(accountAddress: AccountAddress): Proto.AccountAddress {
...
 }
+
+/**
+ * Takes a JSON string and converts it to instance of type {@linkcode AccountAddress}.
+ *
+ * @param {TypedJson} json - The typed JSON to convert.
+ * @throws {TypedJsonParseError} - If unexpected JSON string is passed.
+ * @returns {AccountAddress} The parsed instance.
+ */
+export const fromTypedJSON = makeFromTypedJson(JSON_DISCRIMINATOR, fromBase58);
```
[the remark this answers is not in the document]
soerenbf: Didn't know this!

packages/sdk/src/types/util.ts
```diff
@@ -0,0 +1,163 @@
...
+ * @template D - The JSON discriminator
+ * @template V - The serializable JSON value
+ */
+export interface ToTypedJson<V> {
+    /**
+     * Converts type to {@linkcode TypedJson}
+     *
+     * @returns {TypedJson} The typed JSON.
+     */
+    toTypedJSON(): TypedJson<V>;
```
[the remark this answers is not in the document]
soerenbf: I ran into an issue where these types would be stringified before being sent to WASM, so that's the main reason. Also if users were to stringify types as part of f.x. http requests, I think it would be nice if they didn't have to deal with the discriminator stuff added here.

Will add a comment.

packages/sdk/src/types/json.ts
```diff
@@ -0,0 +1,182 @@
...
+ * @param value A JavaScript value, usually an object or array, to be converted.
+ * @param replacer An array of strings and numbers that acts as an approved list for selecting the object properties that will be stringified.
+ * @param space Adds indentation, white space, and line break characters to the return-value JSON text to make it easier to read.
+ */
+export function jsonStringify(
+    value: any,
+    replacer?: (number | string)[] | null,
+    space?: string | number
+): string;
+export function jsonStringify(
```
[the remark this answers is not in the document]
soerenbf: It was a little bit of a hard exercise for me to do this in an immutable way, but I have something that works now.

I don't know if it's wasted effort though, the recursion limit seems to be very high (>10.000), and I would be surprised if anyone uses JSON objects with anywhere near that level of nesting (although of course it is _theoretically_ possible..

packages/sdk/src/types/json.ts
```diff
@@ -0,0 +1,182 @@
...
+    input: any,
+    replacer?: any,
+    space?: string | number
+): string {
+    /**
+     * Recursively maps concordium domain types to values that can be revived into their original types.
```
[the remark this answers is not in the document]
soerenbf: Added a check for circular references

packages/sdk/src/types/json.ts
```diff
@@ -0,0 +1,329 @@
...
+
+    type StackItem = {
+        obj: Record<string, any>;
+        path: string[];
+    };
+
+    const checkCircular = getCheckCircular();
+
+    const stack: StackItem[] = [{ obj, path: [] }];
+    const result: Record<string, any> = { ...obj }; // Shallow clone to aviod modifying original.
```
limemloh, line 247: You never read from this before assigning the new values, so no need to shallow clone
```suggestion
    const result: Record<string, any> = {};
```

limemloh (approved):
Looks good

packages/sdk/src/types/json.ts
```diff
@@ -0,0 +1,329 @@
...
+            checkCircular.call(obj, k, originalValue); // Throws if a circular reference is found.
+
+            // Transform concordium types first.
+            const { transformed, value } =
+                transformConcordiumType(originalValue);
+            // Then run values through user defined replacer function.
+            const jsonValue =
+                (value as any).toJSON?.() ??
+                replacer?.call(obj, k, value) ??
+                value;
```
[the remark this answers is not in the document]
soerenbf: No you're right, good catch. The value returned from `toJSON` should be propagated to the replacer function along with the key.

PR #280, opened 2023-10-05
Use ModuleReference in GetModuleList

limemloh:
## Purpose

Fix left over hex string by using `ModuleReference` in `GetModuleList`.


soerenbf (approved):

soerenbf:
Typecheck examples fails, so that needs fixing at least 😉

PR #282, opened 2023-10-09
Add Documentation for ccd-js-gen package

limemloh:
## Purpose

Add documentation of how to use the `ccd-js-gen` package.

This can be review already, but a single TODO is still there to be fixed related to metadata for invoking/dry-running contract calls.

### DIFF
--- a/packages/ccd-js-gen/README.md
+++ b/packages/ccd-js-gen/README.md
@@ -1,0 +1,484 @@
+# Smart Contract Client Generator <!-- omit in toc -->
+
+[![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.0-4baaaa.svg)](https://github.com/Concordium/.github/blob/main/.github/CODE_OF_CONDUCT.md)
+
+Generate TypeScript/JavaScript code for interating with smart contracts and modules on the Concordium blockchain.
+
+- Functions for dry-running and calling entrypoints of the smart contract.
+- Functions for constructing parameters.
+- Parsing logged events, return values and error messages.
+- Structured types for parameter, logged events, return values and error messages.
+
+The code is generated from deployable smart contract modules, meaning it can be done with any smart contract available locally and any smart contract deployed on chain.
+
+### Example usage of a generated client
+
+```typescript
+import * as SDK from "@concordium/node-sdk"; // or use @concordium/web-sdk.
+import * as MyContract from "./generated/my-contract.js"; // Code generated from a smart contract module.
+
+const grpcClient = // Setup gRPC client code here ...
+const contractAddress = SDK.ContractAddress.create(42) // Address of the smart contract.
+const signer = // Keys for signing an account transaction.
+const sender = // AccountAddress
+
+// Create a client for 'my-contract' smart contract.
+const contractClient = await MyContract.create(
+    grpcClient,
+    contractAddress
+);
+
+// Construct the parameter for the 'transfer' entrypoint.
+const receiver = AccountAddress.fromBase58('4UC8o4m8AgTxt5VBFMdLwMCwwJQVJwjesNzW7RPXkACynrULmd');
+const parameter: MyContract.TransferParameter = [{
+    tokenId: "",
+    amount: 10,
+    from: { type: 'Account', content: sender },
+    to: { type: 'Account', content: receiver },
+    data: "",
+}];
+
+// Send transaction for invoking the 'transfer' entrypoint of the smart contract.
+const transactionHash = await MyContract.sendTransfer(contractClient, {
+    senderAddress: sender,
+    energy: Energy.create(12000)
+}, parameter, signer);
+```
+
+<!--toc:start-->
+- [Install the package](#install-the-package)
+- [Using the CLI](#using-the-cli)
+  - [Example](#example)
+- [Using the library](#using-the-library)
+  - [Generate from smart contract module file](#generate-from-smart-contract-module-file)
+  - [Generate from smart contract module on chain](#generate-from-smart-contract-module-on-chain)
+- [Using the generated client](#using-the-generated-client)
+  - [Generated module client](#generated-module-client)
+    - [The client type](#the-client-type)
+    - [function `create`](#function-create)
+    - [function `createUnchecked`](#function-createunchecked)
+    - [const `moduleReference`](#const-modulereference)
+    - [function `getModuleSource`](#function-getmodulesource)
+    - [type `<ContractName>Parameter`](#type-contractnameparameter)
+    - [function `instantiate<ContractName>`](#function-instantiatecontractname)
+  - [Generated contract client](#generated-contract-client)
+    - [The contract client type](#the-contract-client-type)
+    - [function `create`](#function-create-1)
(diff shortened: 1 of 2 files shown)


packages/ccd-js-gen/README.md
```diff
@@ -0,0 +1,484 @@
+# Smart Contract Client Generator <!-- omit in toc -->
+
+[![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.0-4baaaa.svg)](https://github.com/Concordium/.github/blob/main/.github/CODE_OF_CONDUCT.md)
```
abizjak, line 3: We probably don't need that here, it's sufficient at the top-level.

packages/ccd-js-gen/README.md
```diff
@@ -0,0 +1,484 @@
...
+### Example usage of a generated client
+
+```typescript
+import * as SDK from "@concordium/node-sdk"; // or use @concordium/web-sdk.
+import * as MyContract from "./generated/my-contract.js"; // Code generated from a smart contract module.
+
+const grpcClient = // Setup gRPC client code here ...
+const contractAddress = SDK.ContractAddress.create(42) // Address of the smart contract.
+const signer = // Keys for signing an account transaction.
+const sender = // AccountAddress
```
abizjak, line 23: Wouldn't it be better to show some valid inputs here?

packages/ccd-js-gen/README.md
```diff
@@ -0,0 +1,484 @@
...
+
+- Functions for dry-running and calling entrypoints of the smart contract.
+- Functions for constructing parameters.
+- Parsing logged events, return values and error messages.
+- Structured types for parameter, logged events, return values and error messages.
+
+The code is generated from deployable smart contract modules, meaning it can be done with any smart contract available locally and any smart contract deployed on chain.
+
+### Example usage of a generated client
+
```
abizjak, line 15: Please state here what contract this is generated from, maybe show the entrypoint that you are invoking. Otherwise it's quite hard to understand the transfer example.

packages/ccd-js-gen/README.md
```diff
@@ -0,0 +1,484 @@
...
+For example: generating clients for a smart contract module `my-module` containing
+the smart contracts `my-contract-a` and `my-contract-b`.
+into the directory `./generated` produces the following structure:
+
+```
+generated/
+├─ my-module.js        // Functions for interacting with the 'my-module' smart contract module on chain.
+├─ my-contract-a.js    // Functions for interacting with the 'my-contract-a' smart contract.
+└─ my-contract-b.js    // Functions for interacting with the 'my-contract-b' smart contract.
```
abizjak, line 185: I did not notice that before, but is name conflict handled?

If I call my module `cis2` and it has a single contract `cis2` will that work?

packages/ccd-js-gen/README.md
```diff
@@ -0,0 +1,484 @@
...
+
+- `moduleClient` The client of the on-chain smart contract module.
+- `transactionMetadata` Metadata related to constructing the transaction, such as energy and CCD amount to include.
+- `parameter` Parameter to provide the smart contract module for the instantiation.
+
+  _With schema type:_ If the schema contains type information for the parameter, a type for the parameter is generated and used for this function. The type is name `<ContractName>Parameter` where `<ContractName>` is the smart contract name in Pascal case.
+
+  _Without schema type:_ If no schema information is present, the function uses the generic `Parameter` from `@concordium/web-sdk`.
+
+- `signer` The keys of to use for signing the transaction.
```
abizjak, line 273: "of to"

packages/ccd-js-gen/README.md
```diff
@@ -0,0 +1,484 @@
...
+
+**Returns**: Promise with the hash of the transaction.
+
+```typescript
+const myModule: MyModule.Type = ...; // Generated module client.
+const transactionMeta = { ... }; // Transaction meta information.
+const parameter = ...; // Parameter to pass the smart contract init-function.
+const signer = ...; // The keys to use for signing the transaction.
+const transactionHash = await MyModule.instantiateMyContract(myModule, transactionMeta, parameter, signer);
```
abizjak, line 282: Same as above, it'd be useful to provide valid dummy data here.

packages/ccd-js-gen/README.md
```diff
@@ -0,0 +1,484 @@
...
+
+An example for a smart contract with an entrypoint named `launch-rocket`, the function becomes `dryRunLaunchRocket`.
+
+**Parameters**
+
+The function parameters are:
+
+- `contractClient` The client of the smart contract instance.
+- `invoker` TODO
+- `parameter` Parameter to provide to the smart contract entrypoint.
```
abizjak, line 413: What about amount?

abizjak (approved):

packages/ccd-js-gen/README.md
```diff
@@ -0,0 +1,484 @@
...
+
+**Parameters**
+
+The function parameters are:
+
+- `moduleClient` The client of the on-chain smart contract module.
+- `transactionMetadata` Metadata related to constructing the transaction, such as energy and CCD amount to include.
+- `parameter` Parameter to provide the smart contract module for the instantiation.
+
+  _With schema type:_ If the schema contains type information for the parameter, a type for the parameter is generated and used for this function. The type is name `<ContractName>Parameter` where `<ContractName>` is the smart contract name in Pascal case.
```
soerenbf, line 269:
```suggestion
  _With schema type:_ If the schema contains type information for the parameter, a type for the parameter is generated and used for this function. The type is named `<ContractName>Parameter` where `<ContractName>` is the smart contract name in Pascal case.
```

Or another alternative

packages/ccd-js-gen/README.md
limemloh: I don't think so, I do this to reduce the noise in the example, such that the interesting parts are more visible.
The point of this example is more of a showcase to help the reader understand what this tool is providing.

abizjak: But it's an ad-hoc mixture of things that are defined and things that are not. Like the contract address is defined, but the sender account is not, nor is the grpc-client, and all of these look very much the same complexity wise to me.

packages/ccd-js-gen/README.md
limemloh: I have changed this to include the module name as prefix for the contracts

packages/ccd-js-gen/README.md
limemloh: It now takes metadata object similar to when sending a transaction, this provides a way to provide the amount, invoker and energy.

packages/ccd-js-gen/README.md
limemloh: I have added more to this now

packages/ccd-js-gen/README.md
limemloh: I now provide some dummy values for the transaction metadata, but decided not to do so for `signer` and `parameter`.
For `signer` I find it adding a lot of noise, and for `parameter` it will completely depend on the contract details, which I have added in the comment for parameter.

soerenbf (approved):

Issue #240, opened 2023-08-24
Support GRPC changes for the new consensus


rasmus-kirk:
**Sub-tasks**
- [x] `getBakersRewardPeriod`: gets the bakers and finalizers that are current for the reward period of a given block. [#239](https://github.com/Concordium/concordium-node-sdk-js/pull/239) 
- [ ] `getWinningBakersEpoch`: gets the bakers that won rounds in the given (past) epoch.
- [x] `getBakerEarliestWinTime`: gets the projected earliest time that a baker may be required to produce a block. [#239](https://github.com/Concordium/concordium-node-sdk-js/pull/239) 
- [x] `getBlockCertificates`: get the quorum certificate, timeout certificate and epoch finalization entry present in a block. [#239](https://github.com/Concordium/concordium-node-sdk-js/pull/239) 
- [ ] `getFirstBlockEpoch`: get the block hash of the first finalized block in a given epoch.

Following changes to current endpoints:
- [x] `getPoolInfo` is amended to add the commission rates for the current reward period to the result. [#239](https://github.com/Concordium/concordium-node-sdk-js/pull/239) 

**Related subissues:**
#232 
#234 
#235

PR #283, opened 2023-10-10
Feature/new consensus api extension

rasmus-kirk:
## Purpose

Addresses #240 

Have tested on new version of the node.


packages/common/CHANGELOG.md
```diff
@@ -49,6 +61,7 @@
 ### Fixes
  - `verifyWeb3IdCredentialSignature` now supports dates/timestamp attributes.
  - `canProveAtomicStatement` now supports timestamp attributes, handles undefined attribute value correctly and handles strings correctly for range statements.
+>>>>>>> origin/main
```
soerenbf, line 64:
```suggestion
```

😄

packages/common/src/types.ts
```diff
@@ -45,6 +68,14 @@ export type Round = bigint;
 /** An amount of energy */
 export type Energy = bigint;
 
+/**
+ * The number of chain restarts via a protocol update. An effected
+ * protocol update instruction might not change the protocol version
+ * specified in the previous field, but it always increments the genesis
+ * index.
+ */
+export type GenesisIndex = number;
```
soerenbf, line 77: Theres a bunch of other places where we also have types with a `genesisIndex` field which just use the `number` type. Maybe it would make sense to align these to all use this alias?

soerenbf:
Changed the base branch, by logic from previous comment. This will ensure the changes here are applied as a minor version on top of the current major version, which we'll then merge in to the next major.

packages/common/src/types.ts
rasmus-kirk: Sure, I can have a look monday

PR #285, opened 2023-10-11
Upgrade @noble/ed25519 to avoid implicit use of crypto module

soerenbf:
## Purpose

Upgrade the ed25519 dependency we use, to avoid implicit use of the nodejs `crypto` module (and thus the need to polyfill it).

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -32,7 +32,11 @@ module.exports = {
         'import/no-unresolved': [
             2,
             {
-                ignore: ['@concordium/rust-bindings', 'grpc-api', '^#.+$'],
+                ignore: [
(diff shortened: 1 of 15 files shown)


.eslintrc.cjs
```diff
@@ -32,7 +32,7 @@ module.exports = {
         'import/no-unresolved': [
             2,
             {
-                ignore: ['@concordium/rust-bindings', 'grpc-api'],
+                ignore: ['@concordium/rust-bindings', 'grpc-api', '^#.+$'],
```
limemloh, line 35: It is not obvious why this is ignored, could you add a comment to why this was added?

limemloh (approved):

PR #288, opened 2023-10-13
Add basic unit test for ccd-js-gen

### LINKED ISSUE #289: Test for ccd-js-gen
Changes in the SDK might result in ccd-js-gen producing invalid code, there it would be beneficial to have tests run by the CI for catching some of these changes.

limemloh:
## Purpose

Add unit testing setup for `ccd-js-gen` package and tests.
Closes #289 

## Changes

- The tests will generate some clients, ensure the generated code type checks and that they export the expected functions.
- Update pipeline.yml to run ccd-js-gen tests.
- In the SDK, introduced a function `versionedModuleSourceToBuffer` for serializing a smart contract module source into a buffer for storing on disk.

## Checklist

      hard-to-understand areas.


packages/sdk/src/types/VersionedModuleSource.ts
```diff
@@ -41,6 +42,19 @@ export function versionedModuleSourceFromBuffer(
...
 
+/**
+ * Serialize a versioned module source. Useful when saving to file.
+ * @param {VersionedModuleSource} moduleSource The versioned module source to serialize.
+ * @returns {Uint8Array} Buffer with serialized module source.
+ */
+export function versionedModuleSourceToBuffer(
+    moduleSource: VersionedModuleSource
+): Uint8Array {
+    const versionBytes = encodeWord32(moduleSource.version);
```
abizjak, line 53: Shouldn't this be big-endian? That's how we store it elsewhere (e.g., the .v1 format).

As far as I see this function is little endian
```
export function encodeWord32(value: number, useLittleEndian = false): Buffer {
    if (value > 4294967295 || value < 0 || !Number.isInteger(value)) {
        throw new Error(
            'The input has to be a 32 bit unsigned integer but it was: ' + value
        );
    }
    const arr = new ArrayBuffer(4);
    const view = new DataView(arr);
    view.setUint32(0, value, useLittleEndian);
    return Buffer.from(new Uint8Array(arr));
}
```

Unless I'm looking at the wrong function.

abizjak: At least it does not seem to match `versionedModuleSourceFromBuffer`

packages/ccd-js-gen/test/generate-wCCD.test.ts
```diff
@@ -0,0 +1,77 @@
+import * as Gen from '../src/lib.js';
+import { assertTypeChecks } from './testHelpers.js';
+
+const moduleFileName = 'wCCD';
```
abizjak, line 4: Can you please provide a header for this file explaining what the test is, what it depends on, and how to regenerate it.

Including how the wasm file was generated.

packages/ccd-js-gen/test/generate-test-bench.test.ts
```diff
@@ -0,0 +1,76 @@
+import * as Gen from '../src/lib.js';
+import { assertTypeChecks } from './testHelpers.js';
+
+const outDir = __dirname + '/lib';
+const outModuleFile = `${outDir}/test-bench`;
+const outContractFile = `${outDir}/test-bench_smart_contract_test_bench`;
+
+beforeAll(() => {
+    return Gen.generateContractClientsFromFile(
+        __dirname + '/resources/test-bench.wasm.v1',
```
abizjak, line 10: Can you please provide a header for this file explaining what the test is, what it depends on, and how to regenerate it.

Including how the wasm file was generated.

packages/sdk/src/types/VersionedModuleSource.ts
limemloh: `useLittleEndian` is an optional parameter and defaults to being `false`, so it should be using big-endian.
Also used this to store modules from chain for the tests, so it should work since they are able to parse it.

packages/ccd-js-gen/test/generate-wCCD.test.ts
limemloh: I have added a header now

abizjak (approved):

packages/ccd-js-gen/jest.config.ts
```diff
@@ -0,0 +1,27 @@
...
+    transform: {
+        '^.+\\.[jt]sx?$': [
+            'ts-jest',
+            {
+                useESM: true,
+                tsconfig: 'tsconfig.json',
+            },
+        ],
+    },
+    extensionsToTreatAsEsm: ['.ts'],
```
soerenbf, line 24:
```suggestion
    preset: 'ts-jest/presets/js-with-ts-esm',
    moduleNameMapper: {
        '^(\\.\\.?\\/.+)\\.js$': '$1', // Remap esmodule file extensions
    }
```

This would be a similar, more minimal config (I tinkered with this for the SDK tests on another branch). I assumed here, that you don't depend on `@noble/ed25519`, but if you do then obviously that should still be transformed.

packages/ccd-js-gen/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
     "extends": "../../tsconfig-base.json",
-    "include": ["src/**/*", "package.json"],
+    "include": ["src/**/*", "test/**/*", "package.json", "jest.config.ts"],
     "compilerOptions": {
         "module": "NodeNext",
         "moduleResolution": "NodeNext",
```
soerenbf, line 6: Any reason why we specify "NodeNext" here? In the base config we have "Node16", and I feel like the better option is to "lock the version" to ensure that our code remains compatible with node 16 and above. Right now, there is no difference in the modules produced or the module resolution strategy applied, but that might change.

packages/sdk/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- Introduce function `versionedModuleSourceToBuffer` for serializing a versioned module to a buffer, which can be stored in a file.
+
```
soerenbf, line 8: Should probably move this to the section including changes for 7.0.0 (which has only been released as a release candidate so far).

You can change the version to "Unreleased" as well, as I think it makes more sense to keep it like that until the official release.

soerenbf (approved):
Looks good, with some minor things to consider.

packages/ccd-js-gen/tsconfig.json
limemloh: The reason is subtle, but it stems from importing the `package.json`:
```ts
import packageJson from '../package.json' assert { type: 'json' };
```
For safety reasons, newer versions of NodeJS require adding `assert` when importing anything other than JS.
So to be able to support these newer versions, we need the assert (or workaround the need to import package json).
Version 16 does not support import-assert, so we cannot use `Node16`.

packages/sdk/package.json
```diff
@@ -35,7 +35,7 @@
     },
     "imports": {
         "#ed25519": {
-            "node": "./src/shims/ed25519.node.ts",
+            "node": "./lib/esm/shims/ed25519.node.js",
```
soerenbf, line 38: 🔥

soerenbf (approved):

PR #299, opened 2023-10-24
React native compatibility

soerenbf:
## Purpose

Ensure the SDK can be used from react-native projects.

## Changes

- Add an entrypoint for react-native to rust-bindings that doesn't use WASM
- Favor dependencies that work across all platforms
- Polyfill missing functionality

## Checklist

      hard-to-understand areas.


soerenbf:
@orhoj @limemloh 

This PR looks huge, but in reality many of the lines added is for the example react-native project `examples/reactnative`. Everything in there besides parts of `App.tsx` is boilerplate and can be skipped.

abizjak:
Is there going to be some CI check to make sure that the react-native part keeps working correctly?

soerenbf:
> Is there going to be some CI check to make sure that the react-native part keeps working correctly?

There is a unit test for the example app, but I'll look into whether it's feasible to add a react-native specific pass of the unit tests included in `packages/sdk`

packages/sdk/src/wasm/credentialDeploymentTransactions.ts
```diff
@@ -302,5 +302,5 @@ export async function signCredentialTransaction(
     signingKey: HexString
 ): Promise<HexString> {
     const digest = getCredentialDeploymentSignDigest(credDeployment);
-    return Buffer.from(await ed.signAsync(digest, signingKey)).toString('hex');
+    return Buffer.from(ed.sign(digest, signingKey)).toString('hex');
```
limemloh, line 305: **Minor thing:** Why use sync here? It seems to make very little sense when the function is still async

packages/sdk/package.json
```diff
@@ -74,20 +92,30 @@
...
         "@scure/bip39": "^1.2.1",
+        "@stardazed/streams-polyfill": "^2.4.0",
         "@types/big.js": "^6.2.0",
         "big.js": "^6.2.0",
         "bs58check": "^3.0.1",
         "buffer": "^6.0.3",
         "hash.js": "^1.1.7",
         "iso-3166-1": "^2.1.1",
         "json-bigint": "^1.0.0",
+        "react-native-polyfill-globals": "^3.1.0",
```
limemloh, line 112: **Suggestion:** That is a lot of new dependencies just needed for one platform. I would suggest we instruct the user how to polyfill their environment instead of making it part of the library.

packages/sdk/webpack.config.ts
```diff
@@ -6,48 +6,81 @@ import url from 'url';
...
-        extensionAlias: {
-            '.js': ['.ts', '.js'],
+function configFor(
+    target: 'web' | 'node' | 'react-native'
+): webpack.Configuration {
+    const t = target === 'react-native' ? 'web' : target;
+    const entry =
+        target === 'react-native'
+            ? [
+                  resolve(__dirname, './polyfill/react-native.ts'),
```
limemloh, line 16: **Suggestion:** Separate the polyfills into a subpath.

I believe it is bad practice for a library to silently enable polyfills in your environment and we should at least split this into a separate file and instruct the user can enable it for react native.

packages/sdk/package.json
```diff
@@ -8,35 +8,50 @@
     "repository": {
         "type": "git",
         "url": "https://github.com/Concordium/concordium-node-sdk-js",
-        "directory": "packages/common"
+        "directory": "packages/sdk"
     },
     "sideEffects": false,
```
limemloh, line 13: Since we are enabling polyfills for the user just by importing the package, this is not without sideeffects anymore. Not sure if the bundlers for react native respect this field

limemloh (changes requested):

packages/sdk/package.json
```diff
@@ -8,35 +8,50 @@
...
+        "./wasm": {
+            "types": "./lib/esm/pub/wasm.d.ts",
+            "bun": "./src/pub/wasm.ts",
+            "react-native": null,
+            "default": "./lib/esm/pub/wasm.js"
         },
         "./*": {
             "types": "./lib/esm/pub/*.d.ts",
             "bun": "./src/pub/*.ts",
+            "react-native": "./lib/umd/concordium.react-native.min.js",
```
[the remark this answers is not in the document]
soerenbf: While the metro bundler supports package exports under a unstable flag (`unstable_enablePackageExports`), they have decided _not_ to commit to supporting package "imports" for now.

These 2 facts together is the deciding factor for doing it this way. That and also being able to include various polyfills in the bundle to make installation more seamless for the user.

packages/sdk/package.json
```diff
@@ -8,35 +8,50 @@
...
+    "main": "lib/umd/concordium.node.min.js",
+    "browser": "lib/umd/concordium.web.min.js",
+    "react-native": "lib/umd/concordium.react-native.min.js",
+    "cdn": "lib/umd/concordium.web.min.js",
+    "types": "lib/esm/index.d.ts",
     "exports": {
         ".": {
             "types": "./lib/esm/index.d.ts",
             "bun": "./src/index.ts",
+            "react-native": "./lib/umd/concordium.react-native.min.js",
```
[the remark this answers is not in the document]
soerenbf: Same as previous comment.

packages/sdk/src/wasm/credentialDeploymentTransactions.ts
soerenbf: To avoid breaking changes. Async functions use webcrypto under the hood, which is not supported on react native. My first attempt included polyfilling webcrypto, but that was not straightforward.

packages/sdk/package.json
soerenbf: I dont know, but we're exporting a UMD, which cannot be tree-shaken anyway. I can mark the react-native UMD as having side effects though.

packages/sdk/webpack.config.ts
soerenbf: Resolving subpath exports is not stable in metro. I guess we could instruct users to do either

```ts
import "@concordium/web-sdk/react-native-compat"; // subpath export
import "@concordium/web-sdk/lib/esm/pub/react-native-compat"; // deep import
```

Support for subpath export is expected to be stable in [release 0.73 in Q3 2023](https://reactnative.dev/blog/2023/06/21/package-exports-support#the-future-stable-exports-enabled-by-default), so I guess it would be fine to work towards this.

packages/sdk/package.json
soerenbf: Removed them

packages/sdk/webpack.config.ts
soerenbf: Will push polyfill responsibility to consumer

packages/sdk/package.json
limemloh: They seem to still be there though. Am I missing something?

packages/sdk/CHANGELOG.md
```diff
@@ -2,7 +2,11 @@
 
 ## Unreleased
 
-### Fix
+### Added
+
+- Support for using the SDK in a react native environment. Requires installing `react-native-get-random-values` peer dependency in consuming package.
```
limemloh, line 7: I guess this is slightly outdated now, since it requires more than that from the users, maybe just refer to the guide

limemloh (approved):

PR #301, opened 2023-10-26
Expose type BlockItem again

limemloh:
## Purpose

Export type `BlockItem` in the public API again, this was removed accidentally in `web-sdk v7.0.0`.

## Checklist

      hard-to-understand areas.


soerenbf (approved):
Should we move the type to `types.ts` instead? Importing it will result in importing the wallet wasm module as well. I realize the only place this is referenced within the SDK, is from the module it is exported, but I reckon it would be better to be on the safe side?

PR #306, opened 2023-11-09
Add function for stringifying wrapper types as unwrapped, serializable types

soerenbf:
## Purpose

Unwrap types to a representation of the inner value, to make it easier to stringify structures using the newly added types into something sensible (f.x. for http requests).

Types that already had `toJSON` functions on them are kept as is.

## Changes

- Add `toJSON` functions where needed.

## Checklist

      hard-to-understand areas.


soerenbf:
Kept as draft until it's been verified as compatible in https://github.com/Concordium/concordium-dapp-libraries/pull/51

packages/sdk/CHANGELOG.md
```diff
@@ -17,7 +23,7 @@
 ### Added
 
 - Support for using the SDK in a react native environment. Requires polyfilling functionality used within the SDK.
-  Please refer to the guide found at [the react native compatibility guide](https://developer.concordium.software/concordium-node-sdk-js/pages/misc-pages/react-native.html)
+    Please refer to the guide found at [the react native compatibility guide](https://developer.concordium.software/concordium-node-sdk-js/pages/misc-pages/react-native.html)
```
shjortConcordium, line 26: :thinking:

packages/sdk/src/types/ContractAddress.ts
```diff
@@ -21,6 +23,10 @@ class ContractAddress {
         /** The subindex of the smart contract address. */
         public readonly subindex: bigint
     ) {}
+
+    public toJSON(): ContractAddressLike<bigint> {
+        return { index: this.index, subindex: this.subindex };
+    }
```
shjortConcordium, line 29: It's a bit scuffed that toJSON is not compatible with stringify, but I guess there's no way around it... :smiling_face_with_tear:

packages/sdk/test/ci/types/native-json.test.ts
```diff
@@ -0,0 +1,261 @@
...
+        expect(JSONBig.stringify(t)).toEqual(e);
+    });
+});
+
+describe('ModuleReference', () => {
+    test('Serializes as expected', () => {
+        const v =
+            '5d99b6dfa7ba9dc0cac8626754985500d51d6d06829210748b3fd24fa30cde4a';
+        const t = ModuleReference.fromHexString(v);
+        const e = `"00000020${v}"`; // is prefixed with 4 byte length, for some reason
```
shjortConcordium, line 225:
```suggestion
        const e = `"00000020${v}"`; // is prefixed with 4 byte length
```
I don't think a comment on a test is the place to question the length of the length :confused:

shjortConcordium (commented):
These changes do break with the browser wallet, using this with wCCD, gives me
`Error: The given transaction is not valid due to: TypeError: Cannot read properties of undefined (reading 'split').
`
I'll check wallet connect tomorrow.

packages/sdk/test/ci/types/native-json.test.ts
soerenbf: Well... I think you're right 😄

packages/sdk/src/types/ContractAddress.ts
soerenbf: Yeah, well it doesn't work with the browser wallet anyway, so probably have to scrap this anyway.

soerenbf:
> These changes do break with the browser wallet, using this with wCCD, gives me `Error: The given transaction is not valid due to: TypeError: Cannot read properties of undefined (reading 'split'). ` I'll check wallet connect tomorrow.

Upon checking again it seems you're right.. Will need to figure out an alternative

packages/sdk/src/types/json.ts
```diff
@@ -161,3 +222,50 @@ export function jsonStringify(
...
+ * - `'number'`: uses 'json-bigint to safely serialize,
+ * - `'string'`: converts `bigint` to strings
+ * - `undefined`: must be taken care of manually, e.g. in replacer function.
+ * Defaults to 'number'
+ * @param replacer A function that transforms the results.
+ * @param space Adds indentation, white space, and line break characters to the return-value JSON text to make it easier to read.
+ */
+export function jsonUnwrapStringify(
+    input: any,
+    bigintFormat: 'number' | 'string' | undefined = 'number',
```
bisgardo, line 241: Consider making this an enum to make it clearer at the call site what it means.

bisgardo: But I'm a little concerned that this parameter is needed in the first place. It feels brittle. Shouldn't it be decided per type how its fields are serialized?

soerenbf: Well we need to do something about bigints.. in the Wallet connectors we use Numbers, which is unsafe for very Big integers, so this is a great alternative to that at least.

Besides you can still pass a replacer function to be called for individual types.

I'll add some unit tests to see what happens if you pass a replacer function while specifying a bigint format, to make sure it's consistent.

We can also just use json-bigint in the wallet connectors lib, if you're not a fan of including it as an option here. I just thought it'd be convenient.

packages/sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@concordium/web-sdk",
-    "version": "7.0.3",
+    "version": "7.0.4-rc.3",
```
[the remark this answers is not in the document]
soerenbf: Yea I agree. Will make sure the final bump is a minor.

packages/sdk/src/types/json.ts
soerenbf: Changed default to run no default conversion.

shjortConcordium (approved):

Issue #308, opened 2023-11-14
Add `toJSON` and `toString` methods on newtype classes (`sdk/src/types/*`)


soerenbf:
**Task description**
The purpose of adding the functionality above is to:

- Make it easier to use the types in string interpolation (by adding `toString`)
- Standardise how the types are represented in a JSON compatible manner (by adding `toJSON`), i.e.
  - types wrapping single values are unwrapped (e.g. `AccountAddress` and `Timestamp`)
  - types wrapping multiple values are represented as objects including just those values (e.g. `ContractAddress`)
  - bigints should still be represented as bigints, and should be left up to dependants to use something like [`json-bigint`](https://www.npmjs.com/package/json-bigint) or a custom replacer function to tailor how bigints are serialised for their particular use-case.

_IMPORTANT_:
This would not normally be considered a breaking change, but until a browser wallet version with the changes proposed in #307 are implemented and utilised in a released version of the browser wallet, it will be due to the browser wallet currently relying on `instanceof` checks included in [`transformConcordiumType`](https://github.com/Concordium/concordium-node-sdk-js/blob/1c859a070351507f9cbb6ccd06952e95b8711f7c/packages/sdk/src/types/json.ts#L91).

abizjak:
I don't think `toString` and `toJSON` should be the same. For example, we use the `<index, subindex>` format when representing account addresses, but that is not how we do it in JSON.

If you want to name things differently that's fine, but there should be a way to render the account address as `<index, subindex>` string.

soerenbf:
> I don't think `toString` and `toJSON` should be the same. For example, we use the `<index, subindex>` format when representing account addresses, but that is not how we do it in JSON.
> 
> If you want to name things differently that's fine, but there should be a way to render the account address as `<index, subindex>` string.

Adding this for reference, since this has already been discussed:

The purpose of the task is to add
- `toString` method to more easily use the types in contexts that calls `toString` method on types given (e.g. in [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)
- `toJSON` method to unwrap newtype patterns to their inner values (e.g. `Energy {value: bigint}` = `bigint`, and `ContractAddress {index: bigint, subindex: bigint}` = `{index: bigint, subindex: bigint}`). The latter example might not seem relevant, but the default `toJSON` of classes also includes private fields in the object returned which we cannot have.

lassemoldrup:
[...]

Issue #307, opened 2023-11-13
Add functionality to JSON encode/decode transaction payloads


soerenbf:
Our wallet API's (`@concordium/browser-wallet-api-helpers`, `@concordium/wallet-connectors`) currently rely on converting transaction payloads through `JSON.stringify` in a way that utilises recursively replacing values (e.g. bigints and newtype pattern types) with values which can be serialised and dynamically restored (in the case of the browser wallet).

With the newtype pattern types (`sdk/src/types/*`) introduced in version 7 of the SDK, it's easy to introduce breaking changes unintentionally by f.x. adding a function to the newtype class, which might not be an obvious breaking change to the developer. Currently, though, this would break the browser wallet as the API injected into the browser uses one version of the SDK, whereas a dApp integrating with the wallet might use another version of the SDK where the types do not match.

To avoid this, we should add functionality for JSON encoding/decoding transaction payloads and use this functionality in the communication across channels that require serialilisability of values, e.g. http requests for wallet connect or chrome API for communicating between browser extension contexts.

A natural fit for this added functionality could be the `AccountTransactionHandler` interface in `sdk/src/accountTransactions.ts`.

Consequently, the [browser wallet](https://github.com/Concordium/concordium-browser-wallet) and the [wallet connectors API](https://github.com/Concordium/concordium-dapp-libraries/blob/main/packages/wallet-connectors/src/WalletConnect.ts#L69) should be updated to utilise this.

We should also, if possible (i.e. look into if this is used in the browser wallet), deprecate the functionality exported from `sdk/src/types/json.ts`, as this should be replaced with the added functionality in the dependants.

PR #309, opened 2023-11-16
Fix errors in the GenericContract.

abizjak:
This changes the errors from the useless

```
Instance [object Object] have contract name '[object Object]' on chain. The client expected: '[object Object]'.
```

to

```
Error: Instance <3, 0> has contract name 'CIS2-wCCD' on chain. The client expected: 'credential_registry'.
```

## Purpose

_Describe the purpose of the pull request, link to issue describing the problem, etc.

## Changes

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/CHANGELOG.md
+++ b/packages/sdk/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## Unreleased changes
+
+### Added
+
+- `ContractAddress.toString` function that converts the address to a string in
+  the `<index, subindex>` format.
+  
+### Fixed
+
+- Error messages in `GenericContract` now display the data, e.g., the contract
+  address, rather than `[object Object]`.
+
+
 ## 7.1.0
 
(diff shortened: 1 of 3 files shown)


soerenbf (approved):
Looks good.

There is a task (#308) to make this easier to do for users for these types in the backlog, which will deprecate what you've added (but will break the browser wallet until we don't rely on JSON.stringify (#307) to replace the types with serializable versions).

limemloh (approved):

abizjak:
> Looks good.
> 
> There is a task (#308) to make this easier to do for users for these types in the backlog, which will deprecate what you've added (but will break the browser wallet until we don't rely on JSON.stringify (#307) to replace the types with serializable versions).

I commented in the linked issue, but I don't think we want to show the account address in JSON format here. We use the `<index, subindex>` format when displaying addresses, quite extensively. It's less verbose.

PR #315, opened 2023-11-30
Fix generated ps1 and cmd files

Bargsteen:
## Purpose

`ccd-js-gen` has an executable script that is used by `npx` but soon also the CCD VS Code extension (https://github.com/Concordium/concordium-smart-contract-tools/pull/140).
On Windows `ps1` and `cmd` files are generated automatically based on the file `bin/ccd-js-gen.js`, but the generator mistakes the `-S` argument for `env` to be the program to run instead of `node`. This renders the scripts broken on Windows.
To get around it, I've changed the script so that it calls node without `--no-warnings` (and thus `-S`), but then it spawns a child process of `node --no-warnings` which executes the actual cli file.

This is the generated `ps1` file *prior to* this PR:
```pwsh
#!/usr/bin/env pwsh
$basedir=Split-Path $MyInvocation.MyCommand.Definition -Parent
​
$exe=""
if ($PSVersionTable.PSVersion -lt "6.0" -or $IsWindows) {
  # Fix case when both the Windows and Linux builds of Node
  # are installed in the same directory
  $exe=".exe"
}
$ret=0
if (Test-Path "$basedir/-S$exe") {
  & "$basedir/-S$exe" node --no-warnings "$basedir/../@concordium/ccd-js-gen/bin/ccd-js-gen.js" $args
  $ret=$LASTEXITCODE
} else {
  & "-S$exe" node --no-warnings "$basedir/../@concordium/ccd-js-gen/bin/ccd-js-gen.js" $args
  $ret=$LASTEXITCODE
}
exit $ret
```

The first attempt was to simply remove the `-S` argument for `env`, but that doesn't work on Linux.
This is the second attempt ;)

## Changes

- Modify the `ccd-js-gen.js` script.

## Checklist

      hard-to-understand areas.


limemloh:
The `-S` was needed for supporting linux, but I guess this was not a universal solution.

Bargsteen:
@limemloh Could you have a look at this again? Thanks 😊

limemloh (approved):
Works on linux at least

PR #319, opened 2023-12-13
Add fromJSON and toJSON for transaction payloads

lassemoldrup:
## Purpose

Addresses #307.

## Changes

Extends the `AccountTransactionHandler` interface with `fromJSON` and `toJSON` methods for converting transactions from and to JSON-serializable types.

## Checklist

      hard-to-understand areas.


packages/sdk/src/accountTransactions.ts
```diff
@@ -159,10 +254,45 @@ export class InitContractHandler
...
+            ),
+        };
+    }
+}
+
+interface UpdateContractPayloadJSON {
+    amount: bigint;
+    address: ContractAddress.SchemaValue;
+    receiveName: string;
+    message: string;
```
soerenbf, line 286:
```suggestion
    message: HexString;
```

packages/sdk/src/accountTransactions.ts
```diff
@@ -244,10 +397,26 @@ export class UpdateCredentialsHandler
     deserialize(): UpdateCredentialsPayload {
         throw new Error('deserialize not supported');
     }
+
+    toJSON(
+        updateCredentials: UpdateCredentialsPayload
+    ): UpdateCredentialsPayload {
+        // UpdateCredentialsPayload is already fully JSON serializable, so we just need to clone it.
```
soerenbf, line 404: Comment seems incorrect since we don't clone?

packages/sdk/src/accountTransactions.ts
```diff
@@ -244,10 +397,26 @@ export class UpdateCredentialsHandler
...
+        return updateCredentials;
+    }
+
+    fromJSON(json: UpdateCredentialsPayload): UpdateCredentialsPayload {
+        return json;
+    }
+}
+
+interface RegisterDataPayloadJSON {
+    data: string;
```
soerenbf, line 414:
```suggestion
    data: HexString;
```

packages/sdk/src/accountTransactions.ts
```diff
@@ -98,10 +143,37 @@ export class SimpleTransferWithMemoHandler
...
+        };
+    }
+
+    fromJSON(
+        json: SimpleTransferWithMemoPayloadJSON
+    ): SimpleTransferWithMemoPayload {
+        return {
+            toAddress: AccountAddress.fromBase58(json.toAddress),
+            // The first 2 bytes are the length of the memo buffer, so we need to remove them.
+            memo: new DataBlob(encodeHexString(json.memo.substring(4))),
```
soerenbf, line 163: I think it would be suitable to define a helper function for this inverse of `DataBlob.toJSON` in the `types/DataBlob.ts` module to keep the transformation logic in one place. And of course use that where applicable in this file 😄

packages/sdk/src/types.ts
```diff
@@ -792,7 +792,7 @@ export interface ChainArData {
 export interface Policy {
     validTo: string; // "YYYYMM"
     createdAt: string; // "YYYYMM"
-    revealedAttributes: Record<AttributeKey, string>;
+    revealedAttributes: Partial<Record<AttributeKey, string>>;
```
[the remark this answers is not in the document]
lassemoldrup: I changed it because my IDE was complaining 🙂

Your test checks if `one` and `two` satisfies _either_ `First` or `Second`, so it passes because it satisfies `Second`. If you remove `| Second` it no longer works.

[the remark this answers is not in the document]
soerenbf: You're right, sorry for that 😜. The previous version is even incorrect, so I'll eat my own hat 🎩. Good find!

packages/sdk/src/types.ts
```diff
@@ -818,7 +818,7 @@ export interface CredentialDeploymentCommitments {
     cmmPrf: string;
     cmmCredCounter: string;
     cmmIdCredSecSharingCoeff: string[];
-    cmmAttributes: Record<AttributeKey, string>;
+    cmmAttributes: Partial<Record<AttributeKey, string>>;
```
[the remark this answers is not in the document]
soerenbf: This change is a more correct model of what is actually being deployed, so we'll keep this. Good find 😄

packages/sdk/src/types/DataBlob.ts
```diff
@@ -30,10 +33,15 @@ export class DataBlob {
         this.data = Buffer.from(data);
     }
 
-    public toJSON(): string {
+    public toJSON(): HexString {
         return packBufferWithWord16Length(this.data).toString('hex');
     }
 
+    public static fromJSON(value: HexString): DataBlob {
+        // The first 2 bytes are the length of the data buffer, so we need to remove them.
```
soerenbf, line 41: Would be good to have this comment as a doc-comment for the function instead to provide documentation for users. Would make sense to describe that it parses the format of `DataBlob.toJSON`

soerenbf (approved):
Looks good 😄

PR #320, opened 2023-12-19
Use concordium-base wallet library

orhoj:
## Purpose
Use `wallet_library` from concordium-base to create identity requests, identity recover requests and for creating unsigned credentials.

This also comes with the addition of two new methods for creating these:
- Creating an identity request without providing the whole seed,
- Creating an identity recovery request without providing the whole seed.

## Changes
- Remove custom implementations from `rust-bindings` and depend on `wallet_library` in concordium-base instead.
- Expose three new WASM entrypoints (`createIdRequestWithKeysV1`, `createIdentityRecoveryRequestWithKeys` and `createUnsignedCredentialWithKeysV1`) from `rust-bindings`. Updated SDK accordingly to provide access to this functionality.
- Change WASM entrypoint `createUnsignedCredentialV1` to use the seed instead of being the function that takes the secret key material directly.
- Update `example/wallet` to use identity creation and unsigned credential creation that does not provide the whole seed to the library.

## Checklist

      hard-to-understand areas.
- [x] (If necessary) I have updated the CHANGELOG. I accept the above linked CLA.


packages/sdk/src/wasm/identity.ts
```diff
@@ -10,47 +10,162 @@ import type {
...
+    net: Network;
+    seedAsHex: string;
+    identityIndex: number;
+};
+
+export type IdentityRequestWithKeysInput = IdentityRequestInputCommon & {
+    prfKey: string;
+    idCredSec: string;
+    blindingRandomness: string;
+};
```
soerenbf, line 37: Would be good if we document at least all the public code we add.

packages/sdk/src/wasm/identity.ts
```diff
@@ -10,47 +10,162 @@ import type {
...
+        blindingRandomness,
+    };
+
+    const rawRequest = wasm.createIdRequestWithKeysV1(
+        JSON.stringify(internalInput)
+    );
+    try {
+        return JSON.parse(rawRequest).idObjectRequest;
+    } catch (e) {
+        throw new Error(rawRequest);
```
soerenbf, line 68: Maybe I'm missing something here, but why you know why we throw an error with the response from the wasm function? I see that this is what we do everywhere, but it seems off to me (unless we use the error case for something somewhere?).

If we do use the error case, perhaps we should add a `@throws` declaration to the jsdoc describing the use case

soerenbf (approved):
Looks good with a nice chunk of code removed 🥇 

As mentioned in one of the comments, it would be good if we could add documentation in the form of jsdoc to at least everything new we export 😄

packages/sdk/src/wasm/identity.ts
soerenbf: I'm fairly sure the error case of the rust `Result` is translated to javascript `throw`, with the concrete error type (`JsError`) returned in the `Result` is translated to a javascript `Error` instance, so in that case this only checks that the successful wasm invocation is parsable by `JSON.parse`.

orhoj: I repeated Mr. Hjorts comment :smile: 

If what you are saying is the case, then the error gotten directly from JSON.parse is probably better to just get directly. Or perhaps add `rawRequest` to the error message gotten there as some more context helps debugging the JSON errors. It usually gives us something like:
```
VM318:1 Uncaught SyntaxError: Unexpected token 'F', "Failure" is not valid JSON
    at JSON.parse (<anonymous>)
    at <anonymous>:1:6
```
and without the given JSON it can sometimes be hard to find exactly where the issue lies. Especially when the JSON is huge.

Can we agree to keep it as is within this PR, and if we want to change it in general, then it should be done separately?

PR #323, opened 2024-01-09
ccd-js-gen: Flag for generating @ts-nocheck

### LINKED ISSUE #313: [ccd-js-gen] Make it possible to skip checking generated files
**Description**
The typescript compiler option `noUnusedLocals` produces compilation errors for the generated modules due to the `private __nominal` fields on the classes.

I think it would make sense to add the option to add a `// @ts-nocheck` line at the top of the file when generating typescript (this is also possible with e.g. the tool we use to generate the GRPC interface).

It might also make sense to support generating es modules when targeting javascript.

limemloh:
## Purpose

Closes #313.
Provide flag for generating `// @ts-nocheck`.

## Changes

- Introduce a new option in both CLI and library for enabling this.
- Extend some details in the changelog.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/ccd-js-gen/src/cli.ts
+++ b/packages/ccd-js-gen/src/cli.ts
@@ -13,7 +13,7 @@ type Options = {
     outDir: string;
     /** The output file types for the generated code */
     outputType: lib.OutputOptions;
-    /** Generate `// @ts-nocheck` annotations in each file. Disabled by default. */
+    /** Generate `// @ts-nocheck` annotations in each file. */
     tsNocheck: boolean;
 };
 
@@ -53,7 +53,7 @@ export async function main(): Promise<void> {
             'Everything'
         )
         .option(
-            '--ts-nocheck',
+            '-n, --ts-nocheck',
             'Generate `@ts-nocheck` annotations at the top of each typescript file.',
             false
         )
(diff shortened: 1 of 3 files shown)


DOBEN:
I still get the error:
```
'__nominal' is declared but its value is never read.
private __nominal = true;
```

It seems the `// @ts-nocheck` needs to be really at the top of the file. Currently, it is not at the top.

    
Screenshot from 2024-01-09 09-38-51

packages/ccd-js-gen/src/cli.ts
```diff
@@ -13,6 +13,8 @@ type Options = {
     outDir: string;
     /** The output file types for the generated code */
     outputType: lib.OutputOptions;
+    /** Generate `// @ts-nocheck` annotations in each file. Disabled by default. */
```
DOBEN, line 16: I would write `False` instead of `Disabled` but not feeling strongly about it.
```suggestion
    /** Generate `// @ts-nocheck` annotations in each file. False by default. */
```

limemloh: For this line, I have just removed it, since it is not optional in the type.

DOBEN (approved):

PR #324, opened 2024-01-10
ccd-js-gen: Generate functions for generating web wallet parameters

### LINKED ISSUE #298: ccd-js-gen: Support signing with web-wallet
The code generated by `ccd-js-gen` is still a bit cumbersome to use when signing using the web-wallet.
Since it requires the parameters in the JSON format of the schema and the schema type itself as the parameter.

Code for constructing these are already generated by `ccd-js-gen` as an intermediate step of the function for generating the parameter, but is not exposed to the user.
We need to factor out these steps into a new function.

limemloh:
## Purpose

Closes #298.

## Changes

- Fix bug in JSON generated for schema type `ULeb128` and `ILeb128`.
- Generate functions for constructing `TypedParameter` found in `@concordium/wallet-connectors`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/ccd-js-gen/test/generate-wCCD.test.ts
+++ b/packages/ccd-js-gen/test/generate-wCCD.test.ts
@@ -91,8 +91,10 @@ describe('Generated contract client file', () => {
         const module = require(`${outContractFile}.js`);
         expect(module.parseEvent).toBeDefined();
         expect(module.createUnwrapParameter).toBeDefined();
+        expect(module.createUnwrapParameterWebWallet).toBeDefined();
         expect(module.parseErrorMessageUnwrap).toBeDefined();
         expect(module.createTransferParameter).toBeDefined();
+        expect(module.createTransferParameterWebWallet).toBeDefined();
         expect(module.parseErrorMessageTransfer).toBeDefined();
     });
 });


packages/ccd-js-gen/README.md
```diff
@@ -559,3 +595,37 @@ const invokeResult = await MyContract.dryRunLaunchRocket(myContract, invoker, pa
...
+
+**Returns**: Parameter for updating a contract instance in the format used by [`@concordium/wallet-connector`](https://www.npmjs.com/package/@concordium/wallet-connectors).
+
+```typescript
+// Wallet connection from `@concordium/wallet-connector`.
+const webWalletConnection: WalletConnection = ...;
+// Parameter to pass the smart contract init-function. The structure depends on the contract.
+const parameter: MyContract.LaunchRocketParameter = ...;
+// Convert the parameter into the format expected by the wallet.
+const walletParameter = MyContract.createLaunchRocketParameterWebWallet(parameter);
```
DOBEN, line 623: I had to add a type assertion here. Might be useful to mention. The `ccd-js-gen` tool has no access to `TypedSmartContractParameters` type from the `@concordium/wallet-connector` package (since it is not a dependency of this project). So the types are not identical.

```
const walletParameter = MyContract.createLaunchRocketParameterWebWallet(parameter) as unknown as TypedSmartContractParameters;
```

packages/ccd-js-gen/README.md
```diff
@@ -314,6 +316,40 @@ const parameter = ...;
...
+
+**Returns**: Parameter for initializing a contract instance in the format used by [`@concordium/wallet-connector`](https://www.npmjs.com/package/@concordium/wallet-connectors).
+
+```typescript
+// Wallet connection from `@concordium/wallet-connector`.
+const webWalletConnection: WalletConnection = ...;
+// Parameter to pass the smart contract init-function. The structure depends on the contract.
+const parameter: MyModule.MyContractParameter = ...;
+// Convert the parameter into the format expected by the wallet.
+const walletParameter = MyModule.createMyContractParameterWebWallet(parameter);
```
DOBEN, line 343: I had to add a type assertion here. Might be useful to mention. The `ccd-js-gen` tool has no access to `TypedSmartContractParameters` type from the `@concordium/wallet-connector` package (since it is not a dependency of this project). So the types are not identical.

```
const walletParameter = MyContract.createMyContractParameterWebWallet(parameter) as unknown as TypedSmartContractParameters;
```

packages/ccd-js-gen/src/lib.ts
```diff
@@ -894,23 +1041,59 @@ function generateContractEntrypointCode(
...
+                    {
+                        type: receiveParameterTypeId,
+                        name: parameterId,
+                    },
+                ],
+                statements: [
+                    'return {',
+                    `    parameters: ${createReceiveParameterJsonFnId}(${parameterId}),`,
+                    '    schema: {',
+                    "        type: 'TypeSchema' as const,",
```
DOBEN, line 1069: Do we need the `const` here?

```suggestion
                    "        type: 'TypeSchema' ",
```

packages/ccd-js-gen/src/lib.ts
```diff
@@ -498,20 +551,59 @@ function generactionModuleContractCode(
...
+                    {
+                        type: initParameterTypeId,
+                        name: parameterId,
+                    },
+                ],
+                statements: [
+                    'return {',
+                    `    parameters: ${createInitParameterJsonFnId}(${parameterId}),`,
+                    '    schema: {',
+                    "        type: 'TypeSchema' as const,",
```
DOBEN, line 579: Do we need the `const` here?

DOBEN (approved):

packages/ccd-js-gen/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- Generate `create<ContractName>ParameterWebWallet` and `create<EntrypointName>ParameterWebWallet` functions for constructing the smart contract parameters in the format used by the Concordium Web-Wallet.
+- Minor breaking change: `create<EntrypointName>Parameter` function for contract entries and `create<ContractName>Parameter` function in module are no longer generated when parameter schema type is `Unit`.
+  This function is mainly used internally, but were still exposed.
```
DOBEN, line 7: Which function?

packages/ccd-js-gen/README.md
limemloh: It should not complain even without access to the type directly, since Typescript is a structural typing system. So that (should) mean that something is actually different in the types... I'll look into this.

packages/ccd-js-gen/src/lib.ts
limemloh: This is needed to avoid typescript inferring `type: 'TypeSchema'` to `string`, which will result in type errors when this is passed as `TypedSmartContractParameters`.
Having `as const` makes it the type `'TypeSchema'`.

packages/ccd-js-gen/README.md
limemloh: The issue seems to be that the wallet does not support using `bigint` as part of smart contract parameters, the generated will use `bigint` for anything that cannot be represented by `number` in JS, so the generated types are currently incompatible.

packages/ccd-js-gen/src/lib.ts
limemloh: See other comment

packages/ccd-js-gen/CHANGELOG.md
limemloh: Right, I have tried to rephrase this a bit

packages/ccd-js-gen/README.md
limemloh: Is tracked in this issue https://github.com/Concordium/concordium-browser-wallet/issues/420

soerenbf (approved):
Looks good 😄 

A side note: I could not run the examples by executing `yarn run-example <path>`. I don't know what has changed, but I had to do `npx tsx <path>` instead. Can you successfully run the examples?

limemloh:
> Looks good 😄
> 
> A side note: I could not run the examples by executing `yarn run-example <path>`. I don't know what has changed, but I had to do `npx tsx <path>` instead. Can you successfully run the examples?

I managed to run all of those that are related to `ccd-js-gen`. Did you make sure to run `yarn run-example ./wCCD/generate.ts --endpoint http://node.testnet.concordium.com:20000` as the first thing?

PR #328, opened 2024-01-16
Recover and save identity in wallet example

shjortConcordium:
## Purpose

- Add Recovery to wallet example.

## Changes

- Added ConfirmIdentity page, which is the new target of the identity provider callback, which waits until confirmation, then saves the identity object and continues to the identity page.
- Added RecoverIdentity page, which performs recovery, saves the identity object and continues to the identity page.
- Identity page no longer fetches the identity, but gets it from storage instead.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/wallet/src/Identity.tsx
+++ b/examples/wallet/src/Identity.tsx
@@ -18,9 +18,9 @@ function DisplayIdentity({ attributes }: { attributes: AttributeList }) {
 export function Identity() {
     const [missingIdentity, setMissingIdentity] = useState<boolean>(false);
     const identity = useMemo<IdentityObjectV1>(() => {
-        const raw = localStorage.getItem(identityObjectKey);
-        if (raw) {
-            return JSON.parse(raw);
+        const identityObjectJson = localStorage.getItem(identityObjectKey);
+        if (identityObjectJson) {
+            return JSON.parse(identityObjectJson);
         } else {
             setMissingIdentity(true);
         }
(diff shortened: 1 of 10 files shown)


examples/wallet/src/Identity.tsx
```diff
@@ -17,31 +16,20 @@ function DisplayIdentity({ attributes }: { attributes: AttributeList }) {
...
-        return (
-            <div>
-                <h3>Identity creation failed</h3>
-                <div>{error}</div>
-            </div>
-        );
-    }
+    const [missingIdentity, setMissingIdentity] = useState<boolean>(false);
+    const identity = useMemo<IdentityObjectV1>(() => {
+        const raw = localStorage.getItem(identityObjectKey);
```
orhoj, line 21:
```suggestion
        const identityObjectJson = localStorage.getItem(identityObjectKey);
```

examples/wallet/src/RecoverIdentity.tsx
```diff
@@ -0,0 +1,158 @@
...
+        // Move them to the initial setup page.
+        navigate('/');
+    }
+
+    if (!dataLoaded) {
+        // Loading identity providers and cryptographic parameters.
+        return null;
+    }
+
+    async function createIdentity() {
```
orhoj, line 62:
```suggestion
    async function recoverIdentity() {
```

orhoj (approved):

PR #336, opened 2024-02-27
Add toString and toJSON methods for `sdk/src/types` classes and deprecate typed JSON functions

### LINKED ISSUE #308: Add `toJSON` and `toString` methods on newtype classes (`sdk/src/types/*`)

lassemoldrup:
## Purpose

Closes #308.

## Changes

Each of the classes receive `toString`, `toJSON`, and `fromJSON` methods. Several functions related to typed JSON that are no longer useful have been deprecated.

## Checklist

      hard-to-understand areas.


packages/sdk/src/types/TransactionExpiry.ts
```diff
@@ -28,9 +27,31 @@ class TransactionExpiry {
...
+     * Get a string representation of the transaction expiry date in seconds since the Unix epoch.
+     * @returns {string} The string representation.
+     */
+    public toString(): string {
+        return this.expiryEpochSeconds.toString();
     }
+
+    /**
+     * Get a JSON-serializable representation of the transaction expiry date.
+     * @throws If the expiry is not representable as a number.
```
soerenbf, line 40: Huh? Is this a mistake?

packages/sdk/src/types/TransactionExpiry.ts
```diff
@@ -28,9 +27,31 @@ class TransactionExpiry {
...
+     * @returns {bigint} The JSON-serializable representation.
+     */
+    public toJSON(): bigint {
+        return this.expiryEpochSeconds;
+    }
```
soerenbf, line 45: I think we should declare this change in the changelog as a bug fix. It's a breaking change, but it was wrong before.

packages/sdk/src/types/CcdAmount.ts
```diff
@@ -31,11 +30,34 @@ class CcdAmount {
...
+    }
+
+    /**
+     * Get a JSON-serializable representation of the CCD amount in micro CCD.
+     * @returns {bigint} The JSON-serializable representation.
+     */
+    public toJSON(): bigint {
+        return this.microCcdAmount;
     }
 }
```
soerenbf, line 50: This is a breaking change, and returning a `string` is the intended value here. It is inconsistent, but we need to keep it as is for now as otherwise deserialization will fail in a number of places.

packages/sdk/src/types/json.ts
```diff
@@ -147,7 +148,7 @@ function unwrapConcordiumType(value: unknown): unknown | undefined {
         case BlockHash.instanceOf(value):
             return BlockHash.toUnwrappedJSON(value as BlockHash.Type);
         case CcdAmount.instanceOf(value):
-            return (value as CcdAmount.Type).toJSON();
```
soerenbf, line 150: And then this should be reverted as well..

packages/sdk/src/types/TransactionExpiry.ts
soerenbf: The reason why this is OK('ish), is because the normal fix for this is to `JSONBig.stringify(...)`, which will give the number representation without losing precision, i.e. the deserializing party will get the same value.

In the case of changing the representation of `CcdAmount` from `string` to `bigint`, `JSONBig.stringify` will produce a `number` instead of `string`, causing deserialization to fail. An example of where this will fail is mobile wallets.

packages/sdk/src/types/TransactionExpiry.ts
lassemoldrup: I think I was at some point serializing as a `number`

packages/sdk/src/types/ContractAddress.ts
```diff
@@ -28,6 +26,34 @@ class ContractAddress implements ContractAddressLike<bigint> {
...
+     */
+    public toString(): string {
+        return `<${this.index}, ${this.subindex}>`;
+    }
+
+    /**
+     * Get a JSON-serializable representation of the contract address.
+     * @returns {ContractAddressLike<bigint>} The JSON-serializable representation.
+     */
+    public toJSON(): ContractAddressLike<bigint> {
```
limemloh, line 42: I think it is odd to return something which is not valid JSON from such a method.
I'll take the discussion in the issue related to this.

packages/sdk/src/types/ContractAddress.ts
```diff
@@ -149,7 +176,8 @@ export function fromSerializable(value: Serializable): ContractAddress {
 }
 
 /**
- * Converts {@linkcode ContractAddress} into {@linkcode Serializable}
+ * Converts {@linkcode ContractAddress} into {@linkcode Serializable}.
+ * @deprecated Use the {@linkcode ContractAddress.toJSON} method instead.
```
limemloh, line 180: Why are we deprecating this? We could have the `toJSON` method call this, and provide both functions with no real downside.
To me, this seems like breaking change without any real value gained.

packages/sdk/src/types/TransactionExpiry.ts
```diff
@@ -28,9 +27,30 @@ class TransactionExpiry {
...
     ) {}
 
-    public toJSON(): number {
-        return Number(this.expiryEpochSeconds);
+    /**
+     * Get a string representation of the transaction expiry date in seconds since the Unix epoch.
+     * @returns {string} The string representation.
+     */
+    public toString(): string {
+        return this.expiryEpochSeconds.toString();
```
limemloh, line 35: **Suggestion:** I think it would be more useful for the user to see a date-time instead

packages/sdk/src/types/ContractAddress.ts
soerenbf: This represents the integers as strings, whereas `toJSON` does bigints, so they're not the same. They were mostly there as a helper function for `toTypedJSON`.

Of course this depends on the resolution of the discussion in https://github.com/Concordium/concordium-node-sdk-js/issues/308#issuecomment-1976696107

packages/sdk/src/types/TransactionExpiry.ts
limemloh: Right, I see the issue

packages/sdk/src/types/TransactionExpiry.ts
```diff
@@ -28,11 +24,32 @@ class TransactionExpiry {
...
+    }
+
+    /**
+     * Get a JSON-serializable representation of the transaction expiry date.
+     * @returns {string} The JSON-serializable representation.
+     */
+    public toJSON(): string {
+        return this.expiryEpochSeconds.toString();
```
soerenbf, line 40: So this used to be serialized as a `number`, but is now a `string`? I don't know if we should do this, as it might have unintended consequences (the output of `JSON.stringify` will change without warning).

I would (again) prefer that we change this to return a `bigint`, which at least brings the error to the user at the point of serializing the type.

@limemloh @lassemoldrup what do you think we should do here?

packages/sdk/src/types/json.ts
```diff
@@ -72,7 +72,8 @@ function reviveConcordiumTypes(value: unknown) {
 }
 
 /**
- * Acts as an inverse for {@linkcode jsonStringify}
+ * Acts as an inverse for {@linkcode jsonStringify}.
+ * @deprecated Use an `AccountTransactionHandler` instance instead.
```
soerenbf, line 76: I guess this is not correct, as this can take any structure of concordium domain types (not just account transaction payloads).
```suggestion
 * @deprecated Manually convert the JSON structure instead. For account transactions, parse the output with something that handles deserializing`bigint`s, e.g. the `json-bigint` dependency, and use `AccountTransactionHandler.fromJSON`.
```

packages/sdk/src/types/json.ts
```diff
@@ -206,6 +209,7 @@ function ccdUnwrapReplacer(this: any, key: string, value: any): any {
 /**
  * Stringify, which ensures concordium domain types are stringified in a restorable fashion.
  * This should be used if you want to be able to restore the concordium domain types in the JSON to its original types.
+ * @deprecated Use an `AccountTransactionHandler` instance instead.
```
soerenbf, line 212:
```suggestion
 * @deprecated Manually convert the object to the preferred JSON structure instead. For account transactions, use `AccountTransactionHandler.toJSON` prior to invoking `JSON.stringify`. It's up to the developer to handle serialization of `bigints`, e.g. with the `json-bigint` dependency.
```

packages/sdk/src/types/json.ts
```diff
@@ -238,6 +242,7 @@ export const enum BigintFormatType {
 /**
  * Stringify, which ensures concordium domain types are unwrapped to their inner type before stringified.
  * This should be used if you want to manually deserialize the inner property values, as the serialization is irreversible.
+ * @deprecated Use an `AccountTransactionHandler` instance instead.
```
soerenbf, line 245: Same as above

packages/sdk/src/types/TransactionExpiry.ts
lassemoldrup: The problem is that converting to a number can fail if the internal `bigint` is too large. We discussed this and decided that converting to a `number` should be considered a bug, thus it was changed.

As to whether `string` or `bigint` is the right type, I think that since this is a one-off case, having a type that you can call `JSON.stringify` on is more important than any compatibility concerns. If you want the `bigint`, it is also easily accessible.

soerenbf: What I mean is that we have existing infrastructure that might depend on this returning a `number` type since it has been declared (as opposed to the many types that had no `toJSON` method).

Taking the pragmatic standpoint, the current (faulty) implementation will not fail as it will be a very long time until `Number.MAX_SAFE_INTEGER` is hit with seconds from unix epoch.

```
new Date(Number.MAX_SAFE_INTEGER) // Invalid date
new Date(Number.MAX_SAFE_INTEGER / 10) // Mon Sep 05 30512 05:17:54 GMT+0200 (Central European Summer Time)
```

I would rather keep the faulty implementation until we can introduce this as a breaking change if we want to 🤷

lassemoldrup: After discussion with @limemloh, we decided to keep `number` and throw an error if the value is not a safe integer.

soerenbf (approved):
Looks good!

PR #352, opened 2024-04-23
CIS-3 support

### LINKED ISSUE #343: Support CIS3 client
**Task description**

Like CIS2 client.

lassemoldrup:
## Purpose

Closes #343.

## Changes

Added `CIS3Contract` class for interacting with contracts adhering to the CIS-3 standard and types and functions for deserializing CIS-3 events.

## Checklist

      hard-to-understand areas.


packages/sdk/src/serialization.ts
```diff
@@ -65,19 +65,27 @@ function serializeAccountTransactionHeader(
...
         const signatureBytes = Buffer.from(signature, 'hex');
-        const length = Buffer.alloc(2);
-        length.writeUInt16BE(signatureBytes.length, 0);
-        return Buffer.concat([length, signatureBytes]);
+        if (serializeSignatureLength) {
+            const length = Buffer.alloc(2);
+            length.writeUInt16BE(signatureBytes.length, 0);
+            return Buffer.concat([length, signatureBytes]);
+        } else {
+            return Buffer.concat([Buffer.from([0]), signatureBytes]);
```
limemloh, line 87: **Request change:** This serialization is very specific to CIS-3, so I don't think it should be part of this function, probably better to have it as a separate function.
Also, the `0` prefix needs to be explained in a comment.

packages/sdk/src/serializationHelpers.ts
```diff
@@ -431,3 +431,10 @@ export const makeSerializeOptional =
 
         return Buffer.concat([encodeBool(true), fun(value)]);
     };
+
+export const makeSerializeList =
```
limemloh, line 435: **Suggestion (non-blocking):** Avoid introducing this function and use `serializeList` from this file instead.

packages/sdk/test/client/CIS3Contract.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    '6fb89ee03ea03d038ba0bfa62dbc29eec6f30b154f797cfa5c296beb1c178428';
+// Contract that supports CIS2 tokens and the use of `permit` for transfers. SPONSOREE has a lot of tokens.
+const CIS2_MULTI_CONTRACT = ContractAddress.create(8657);
+const TOKEN_ID = '01';
+const TEST_BLOCK = BlockHash.fromHexString(
+    '5f11cac7e4b81784dda33d46ffdef94e886db3321fb3a995f6b63eef8d4b68b7'
+);
+
+const in5Minutes = () =>
+    Timestamp.fromDate(new Date(Date.now() + 1000 * 60 * 5));
```
limemloh, line 36: **Suggestion:** Add this as a `futureMinutes` helper in the `Timestamp` module.

We already have it for `TransactionExpiry`.

limemloh (approved):
Looks good to me, only minor comments

packages/sdk/src/serialization.ts
lassemoldrup: Just to be sure I understand: you are suggesting copying the whole function and making a `serializeCIS3AccountTransactionSignature` that prepends the zero byte?

limemloh: yes, I think it would be better to keep the two serializations separate.

PR #355, opened 2024-04-26
Fix dependencies of web wallet example

limemloh:
## Purpose

The web-wallet example is missing some dependencies and importing files outside of its scope.
This PR fixes that.
Closes https://github.com/Concordium/concordium-node-sdk-js/issues/354


examples/wallet/src/util.ts
```diff
@@ -34,7 +34,6 @@ import {
     nodePort,
     walletProxyBaseUrl,
 } from './constants';
-import { filterRecord, mapRecord } from '../../../packages/sdk/lib/esm/util';
```
soerenbf, line 37: Wow... that also seems like a bit of a hack 😲

examples/wallet/package.json
```diff
@@ -17,6 +18,7 @@
         "@vitejs/plugin-react-swc": "^3.4.0",
         "eslint": "^8.50.0",
         "typescript": "^5.2.2",
+        "ts-node": "10.9",
```
soerenbf, line 21: Why is this needed? For vite to understand config files in typescript or?

soerenbf (approved):

examples/wallet/package.json
limemloh: Exactly, didn't want to start a fight with the entire configuration, so I just added it here

examples/wallet/src/util.ts
limemloh: yes, probably some auto-completion was the cause of this

PR #362, opened 2024-05-22
[JSS-101] Estimate smart-contract update energy costs

Ivan-Mahda:
## Purpose

We should provide functions for computing an estimate of energy usage needed for running an update transaction for a smart contract on chain.
The function should take the information needed for the update and use invokeContract to compute the energy cost.

## Changes

Added new method getContractUpdateEnergyCost
Get contract update energy cost
Estimated by calculateEnergyCost, where transactionSpecificCost received from invokeContract used energy

Actual energy can be different from the estimated (about 1 - 5 NRG)

Updated README.md for node-sdk examples

image

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/nodejs/README.md
+++ b/examples/nodejs/README.md
@@ -48,3 +48,7 @@ yarn run-example /path/to/example.ts [opts]
 ```
 
 Where opts are any arguments that the example script takes.
+
+Default endpoint for node is 'localhost:20000', 
+but instead of installing local node,
+can be used testnet node https://node.testnet.concordium.com:20000
--- a/packages/sdk/CHANGELOG.md
+++ b/packages/sdk/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 7.4.1
+
+### Added
+
+- Method (`getContractUpdateEnergyCost`) for estimating energy usage of contract update.
+
 ## 7.4.0
 
 ### Added
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@concordium/web-sdk",
-    "version": "7.4.0",
+    "version": "7.4.1",
     "license": "Apache-2.0",
     "engines": {
         "node": ">=16"
--- a/packages/sdk/src/contractHelpers.ts
+++ b/packages/sdk/src/contractHelpers.ts
@@ -91,3 +91,16 @@ export function getNamesFromReceive(receiveName: string): {
         entrypointName: receiveName.substring(splitPoint + 1),
     };
 }
+
+/**
+ * Get contract update payload size by adding reserved offsets to parameter size and receive name size
+ * Amount (8 bytes), Contract address (16 bytes), Receive name size (2 bytes), Parameter size (2 bytes)
+ */
+export function getUpdatePayloadSize(
+    parameterSize: number,
+    receiveNameLength: number
+) {
+    return (
+        8n + 16n + 2n + BigInt(parameterSize) + 2n + BigInt(receiveNameLength)
+    );
+}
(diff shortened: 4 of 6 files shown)


examples/nodejs/README.md
```diff
@@ -48,3 +48,7 @@ yarn run-example /path/to/example.ts [opts]
 ```
 
 Where opts are any arguments that the example script takes.
+
+Default endpoint for node is 'localhost:20000',
+but instead of installing local node,
+can be used testnet node <https://node.testnet.concordium.com:20000>
```
limemloh, line 54: This endpoint has been replaced
```suggestion
can be used testnet node <https://grpc.testnet.concordium.com:20000>
```

.markdown-linkcheck.json
```diff
@@ -2,6 +2,9 @@
     "ignorePatterns": [
         {
             "pattern": "classes/grpc.ConcordiumGRPCClient.html"
+        },
+        {
+            "pattern": "https://node.testnet.concordium.com:20000"
```
limemloh, line 7: We are migrating to another endpoint
```suggestion
            "pattern": "https://grpc.testnet.concordium.com:20000"
```

packages/sdk/src/energyCost.ts
```diff
@@ -60,6 +68,42 @@ export function getEnergyCost(
...
+/**
+ * Get contract update energy cost
+ * Estimated by calculateEnergyCost, where transactionSpecificCost received from invokeContract used energy
+ * @param {ConcordiumGRPCClient} grpcClient - The client to be used for the query
+ * @param {ContractAddress.Type} contractAddress - The address of the contract to query
+ * @param {AccountAddress.Type} invoker - Representation of an account address
+ * @param {Parameter.Type} parameter - Input for contract function
+ * @param {ReceiveName.Type} method - Represents a receive-function in a smart contract module
+ * @param {bigint} signatureCount - Number of expected signatures
+ */
```
limemloh, line 80: **Suggestion:** Please document when this function throws and also that the estimate is given for the last finalized block according to the node, which means this the actual energy cost might be different depending on the implementation of the smart contract and the interaction with the instance, since this was estimated.

limemloh (approved):

packages/sdk/src/contractHelpers.ts
```diff
@@ -91,3 +91,16 @@ export function getNamesFromReceive(receiveName: string): {
...
 }
+
+/**
+ * Get contract update payload size by adding reserved offsets to parameter size and receive name size
+ * Amount (8 bytes), Contract address (16 bytes), Receive name size (2 bytes), Parameter size (2 bytes)
+ */
+export function getUpdatePayloadSize(
+    parameterSize: number,
+    receiveNameLength: number
+) {
```
soerenbf, line 102:
```suggestion
): bigint {
```

packages/sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@concordium/web-sdk",
-    "version": "7.4.0",
```
soerenbf, line 3: Would argue that this is a minor version (as it adds new functionality). The way we usually release though, is that we create a pull request specifically for bumping the version in preparation for the release. This is to avoid confusion with regards to what has already been released and what is pending release.

packages/sdk/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 7.4.1
```
soerenbf, line 3: Consequently, this should be

```suggestion
## Unreleased
```

packages/sdk/src/energyCost.ts
soerenbf: Would it not make sense to add an optional `blockHash` argument which can be passed to the grpc client?

soerenbf (commented):
Mainly comments regarding how we usually bump/release

PR #366, opened 2024-05-28
Add size limit CI

### LINKED ISSUE #360: Make CI check bundle size
We should set up the CI using \[size-limit\]([https://github.com/ai/size-limit](https://github.com/ai/size-limit)) to ensure that we don't accidentally blow up the bundled size of the SDK.<br>Small changes can result in bad tree-shaking and this would catch some of it.

lassemoldrup:
## Purpose

Closes #360.

## Changes

- Adds `yarn size`, which checks that tree shaking is working when only importing `ConcordiumGRPCWebClient`.
- Adds a step to the pipeline that tests this.

## Checklist

      hard-to-understand areas.


packages/sdk/.size-limit.ts
```diff
@@ -0,0 +1,10 @@
+import type { SizeLimitConfig } from 'size-limit';
+
+module.exports = [
+    {
+        name: "GRPC web client tree shaking",
+        path: "lib/esm/index.js",
+        limit: "100 KB",
```
limemloh, line 7: I think we should set the limit, just above the current size (53.08 kB), such that we catch if the bundle size grows.
So maybe put `54 kB` instead

limemloh (approved):

PR #367, opened 2024-05-31
Merge dapp libs

### LINKED ISSUE #304: Configure prettier to run on its own (i.e. not as an ESLint rule)
The recommended way to run prettier is to run it on its own and not as a plugin of something else (source [https://prettier.io/docs/en/integrating-with-linters.html](https://prettier.io/docs/en/integrating-with-linters.html)). I think we should integrate `prettier --check` as part of our CI instead of running it on lint.

Also prettier by default formats a lot of things aside from js/ts files. Some editors are configured to format files on save, and currently this repository doesn't support that very well, as prettier will run on every file supported by it (unless ignored). This will result in some diffs which consist largely of formatting changes due to being implemented in an editor with this workflow integrated. I would suggest we handle this by:

- [X] Updating prettier to version 3 (ignores files from `.gitignore` and `.prettierignore` by default)
- [X] Remove eslint integration with prettier, add `eslint-config-prettier` instead, which simply disables rules which would conflict with prettier)
- [X] Add `.prettierignore` to select which files we want to format
- [X] Run `prettier --check` along with lint in CI
- [ ] ~Add pre-commit hook to run prettier on staged files of selected file types (to keep functionality as is)~

soerenbf:
## Purpose

Formatting applied to all workspace members in #368. To make CI pass, this PR must be merged in. These are separated to avoid formatting making the actual changes related to merging the repositories impossible to review.

Closes #304

## Checklist

      hard-to-understand areas.


limemloh (approved):
Looks good to me

PR #381, opened 2024-08-07
Fix the rust version used in pipeline

limemloh:
## Purpose

Fix issue causing the rust-bindings wasm module to throw runtime error: `unreachable`.

## Changes

- Use a fixed Rust toolchain version everywhere in the pipeline. This will fix an issue in the deployment, since the wasm-pack version `0.12.1` seems incompatible with the rust version actually used by the CI (which was `1.79` by accident).
- Add unit test catching the above issue, when we update the rust version sometime in the future.
- Correct path for caching of `concordium-contracts-common` workspace.
- Remove unused type `AttributeRandomness`.


soerenbf (approved):
Looks good 👍 😄

PR #384, opened 2024-08-27
Pv7 conformance

### LINKED ISSUE #383: Align `TransactionType` enum with GRPC api types
In the current version of the SDK, transaction types are numbered according to the V1 API specification. Our other SDK's seem to follow the GRPC V2 `TransactionType` in terms of the underlying value of this type.

soerenbf:
## Purpose

Update the SDK to conform to the changes introduced to the node API with protocol version 7

Closes #383

## Checklist

      hard-to-understand areas.


packages/sdk/src/types.ts
```diff
@@ -1187,18 +1234,18 @@ export enum AccountTransactionType {
...
+    TransferToEncrypted = 11,
+    TransferToPublic = 12,
+    TransferWithSchedule = 13,
+    UpdateCredentials = 14,
+    RegisterData = 15,
+    TransferWithMemo = 16,
+    EncryptedAmountTransferWithMemo = 17,
+    TransferWithScheduleAndMemo = 18,
+    ConfigureBaker = 19,
+    ConfigureDelegation = 20,
```
soerenbf, line 1248: These changes make the unit tests fail, as one test tests the expected transaction hash of a transaction. To be honest I have a hard time understanding how the previous values were accepted by the node? The changed values correspond to the values in the rust SDK' and in the GRPC api proto types.

limemloh: According to the doc comment, these values are the ones used for the tag in the serialization, so you should not look at the gRPC types.
The values should match [the tag found here](https://github.com/Concordium/concordium-base/blob/main/rust-src/concordium_base/src/transactions.rs#L1037)

soerenbf: Huh.. I'll revert this then 😅

packages/sdk/src/grpc/translation.ts
```diff
@@ -346,6 +351,8 @@ export function accountInfo(acc: v2.AccountInfo): v1.AccountInfo {
         total: CcdAmount.fromProto(unwrap(acc.schedule?.total)),
         schedule: unwrap(acc.schedule?.schedules).map(trRelease),
     };
+    const cooldowns = acc.cooldowns.map(transCooldown);
+    const availableBalance = CcdAmount.fromProto(unwrap(acc.availableBalance));
```
limemloh, line 355: **Suggestion:** The `unwrap` here will mean this query fails when used against concordium-nodes prior to `7.0.0`.
So we should probably document the minimum support node version somewhere.

packages/sdk/src/grpc/translation.ts
```diff
@@ -830,8 +841,8 @@ function trBakerEvent(bakerEvent: v2.BakerEvent, account: AccountAddress.Type):
                 account,
             };
         }
-        case undefined:
-            throw Error('Failed translating BakerEvent, encountered undefined');
+        default:
+            throw Error('Unrecognized event type. This should be impossible.');
```
limemloh, line 845: **Suggestion (non-blocking):** Won't this mean that we don't get compilation errors for non-exhaustive matching?

packages/sdk/src/types.ts
```diff
@@ -1037,6 +1069,21 @@ interface AccountInfoCommon {
     accountEncryptedAmount: AccountEncryptedAmount;
     accountReleaseSchedule: AccountReleaseSchedule;
     accountCredentials: Record<number, AccountCredential>;
+    /**
+     * The stake on the account that is in cooldown.
+     * There can be multiple amounts in cooldown that expire at different times.
+     * This was introduced in protocol version 7, and so is not present in
+     * earlier protocol versions.
```
limemloh, line 1076: **Suggestion (non-blocking):** The wording makes it sound like it will be undefined prior to P7, but I think a more precise wording would be that: it will always be the empty array for any block validated prior to P7.

limemloh (approved):
Great work adding more documentation to types :pray:

packages/sdk/src/grpc/translation.ts
soerenbf: True.. I didn't actually think that the ts compiler checked if all cases were checked in a switch. I'll try again 👍

packages/sdk/src/types.ts
soerenbf: I copied it from the docs in the GRPC proto types, but you're right. Good catch 😄

packages/sdk/src/grpc/translation.ts
soerenbf: Yes, this will be merged into the release branch for version 8 of the SDK, and I think the strategy should be to release it in stages through npm tags:

1. alpha = local/stagenet node 7.0 release
2. beta = testnet node 7.0 release
3. latest = mainnet node 7.0 release

I'll figure out where to document the minimum supported node version as well.

soerenbf: I added a release schedule to https://github.com/Concordium/concordium-node-sdk-js/pull/385 which is a release PR for the target branch.

packages/sdk/test/client/clientV2.test.ts
soerenbf, line 1: The lookup for the credential ID fails (it also fails when trying to query through the concordium client). The test basically tests the implementation of the node, i.e. looking up accounts by credId/address returns the same result. As such, it provides no real value..

PR #389, opened 2024-09-06
Add fallback for available balance to be compatible with node version 6

soerenbf:
## Changes

- Add fallback for `AccountInfo.accountAvailableBalance`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/src/grpc/translation.ts
+++ b/packages/sdk/src/grpc/translation.ts
@@ -339,7 +339,7 @@ export function accountInfo(acc: v2.AccountInfo): v1.AccountInfo {
     const aggAmount = acc.encryptedBalance?.aggregatedAmount?.value;
     const numAggregated = acc.encryptedBalance?.numAggregated;
 
-    const encryptedAmount: v1.AccountEncryptedAmount = {
+    const accountEncryptedAmount: v1.AccountEncryptedAmount = {
         selfAmount: unwrapValToHex(acc.encryptedBalance?.selfAmount),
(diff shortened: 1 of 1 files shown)


packages/sdk/src/grpc/translation.ts
```diff
@@ -339,33 +339,60 @@ export function accountInfo(acc: v2.AccountInfo): v1.AccountInfo {
...
         total: CcdAmount.fromProto(unwrap(acc.schedule?.total)),
         schedule: unwrap(acc.schedule?.schedules).map(trRelease),
     };
-    const cooldowns = acc.cooldowns.map(transCooldown);
-    const availableBalance = CcdAmount.fromProto(unwrap(acc.availableBalance));
+    const accountCooldowns = acc.cooldowns.map(transCooldown);
+    const accountAmount = CcdAmount.fromProto(unwrap(acc.amount));
+
+    let accountAvailableBalance: CcdAmount.Type;
+    if (acc.availableBalance !== undefined) {
```
limemloh, line 358: **Suggestion:**
It would be good to add a comment on why this can be undefined, since the current gRPC makes it looks like it cannot, so we should add this context.
It should include what node versions this will actually be relevant for

limemloh (approved):

packages/sdk/src/grpc/translation.ts
soerenbf: Good suggestion, will add this

PR #410, opened 2025-04-02
Plt support: add types and translation of protobuf definitions

soerenbf:
## Purpose

Implements the part of PLT support in the SDK pertaining to translating values received from the GRPC api of the corresponding Concordium node.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/src/plt/TokenAmount.ts
+++ b/packages/sdk/src/plt/TokenAmount.ts
@@ -1,4 +1,5 @@
 import { MAX_U32, MAX_U64 } from '../constants.js';
+import type * as Proto from '../grpc-api/v2/concordium/protocol-level-tokens.js';
 
 /**
  * Protocol level token (PLT) amount JSON representation.
@@ -156,4 +157,26 @@ export function create(value: bigint, decimals: number): TokenAmount {
  */
 export function zero(decimals: number): TokenAmount {
     return new TokenAmount(BigInt(0), decimals);
+}
+
+/**
+ * Convert token amount from its protobuf encoding.
+ * @param {Proto.TokenAmount} amount
+ * @returns {Type} The token amount.
+ * @throws {Err} If the value/digits exceeds the maximum allowed or is negative.
+ */
+export function fromProto(amount: Proto.TokenAmount): Type {
+    return create(amount.digits, amount.nrOfDecimals);
+}
+
+/**
+ * Convert token amount into its protobuf encoding.
+ * @param {TokenAmount} amount
+ * @returns {Proto.TokenAmount} The protobuf encoding.
+ */
+export function toProto(amount: Type): Proto.TokenAmount {
+    return {
+        digits: amount.value,
+        nrOfDecimals: amount.decimals,
+    };
 }
--- a/packages/sdk/src/plt/TokenAmount.ts
+++ b/packages/sdk/src/plt/TokenAmount.ts
@@ -1,0 +1,87 @@
+/**
+ * Protocol level token (PLT) amount JSON representation.
+ *
+ * Please note that `bigint` is used to represent the token amount, which is needed for precise representation of large numbers.
+ * As such, extra steps must be taken to serialize and deserialize the token amount.
+ */
+export type JSON = {
+  /** The integer representation of the token amount. */
+  value: bigint;
+  /** The decimals of the token amount, defining the precision at which amounts of the token can be specified. */
+  decimals: number;
+};
+
+/**
+ * Protocol level token (PLT) amount representation.
+ */
+class TokenAmount {
+  /** Having a private field prevents similar structured objects to be considered the same type (similar to nominal typing). */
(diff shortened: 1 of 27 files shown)


packages/sdk/src/plt/TokenEvent.ts
```diff
@@ -0,0 +1,102 @@
+import * as Proto from '../grpc-api/v2/concordium/protocol-level-tokens.js';
+import { HexString } from '../types.js';
+
+export type JSON = HexString;
```
soerenbf, line 4: Maybe we should
```suggestion
export type JSON = any;
```

packages/sdk/src/plt/TokenEvent.ts
```diff
@@ -0,0 +1,102 @@
...
+    public toString(): string {
+        return toHexString(this);
+    }
+
+    /**
+     * Get a JSON-serializable representation of the token event.
+     * @returns {HexString} The JSON-serializable representation.
+     */
+    public toJSON(): HexString {
+        return toHexString(this);
```
soerenbf, line 30: and correspondingly
```suggestion
        return cbor.decode(this.buffer);
```

packages/sdk/src/plt/TokenEvent.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/**
+ * Converts a {@linkcode HexString} to a token event.
+ * @param {HexString} json The JSON representation of the token event.
+ * @returns {TokenEvent} The token event.
+ */
+export function fromJSON(json: HexString): TokenEvent {
+    return fromHexString(json);
+}
```
soerenbf, line 41: And also
```suggestion
export function fromJSON(json: JSON): TokenEvent {
    return fromBuffer(cbor.encode(json));
}
```

packages/sdk/src/plt/TokenModuleState.ts
soerenbf, line 1: Same changes would be applied here

packages/sdk/src/types.ts
```diff
@@ -1124,6 +1147,8 @@ interface AccountInfoCommon {
      * This was introduced with node version 7.0
      */
     accountAvailableBalance: CcdAmount.Type;
+    /** The protocol leveltokens held by the account. */
+    accountTokens: TokenAccountInfo[];
```
soerenbf, line 1151: In the protobuf types this is defined as a list. We could be more precise and use a `Map<TokenId.Type, TokenAccountState>` here. Might be harder to use on the consumer end though without providing much value. athis also assumes that we can rely on the uniqueness of the token ID. being enforced. I think for the sake of simplicity, it's better to keep it as is.

packages/sdk/src/plt/TokenEvent.ts
limemloh: it makes sense to me that the default JSON encoding is a hex string, but it sounds like a good idea to provide the cbor decode functions as well as long as the library can be tree shaken.

packages/sdk/src/plt/TokenModuleReference.ts
```diff
@@ -0,0 +1,169 @@
+import { Buffer } from 'buffer/index.js';
+
+import type * as Proto from '../grpc-api/v2/concordium/protocol-level-tokens.js';
+import type { HexString } from '../types.js';
+
+/**
+ * The number of bytes used to represent a block hash.
```
limemloh, line 7:
```suggestion
 * The number of bytes used to represent a token module reference.
```

packages/sdk/src/types.ts
limemloh: I agree, protobuf have a map type, but this is also just a wrapper around a list and does not rely on any properties of the key, meaning no performance gains, so the user might as well build the map when that is the most useful to work with.

limemloh (approved):
Great work and love the level of documentation :1st_place_medal:

packages/sdk/src/plt/TokenAmount.ts
```diff
@@ -0,0 +1,246 @@
...
+
+/**
+ * Protocol level token (PLT) amount JSON representation.
+ *
+ * Please note that `bigint` is used to represent the token amount, which is needed for precise representation of large numbers.
+ * As such, extra steps must be taken to serialize and deserialize the token amount.
+ */
+export type JSON = {
+    /** The integer representation of the token amount. */
+    value: bigint;
```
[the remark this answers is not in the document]
soerenbf: We decided to convert the `bigint` to `string` here as it will allow parsers to accept both `int` and "stringified `int`", while producing valid JSON without the need of a third party library for bigint conversion on the side of the consumer of the SDK.

packages/sdk/src/plt/TokenEvent.ts
soerenbf: This will be implemented in another PR, which will include general serialization of plt types and transactions.

PR #411, opened 2025-04-08
Plt token client

soerenbf:
## Purpose

Implements token clients to enable users to interact with PLTs, both at a version-agnostic level, and with a specific version (currently only v1).

## Checklist

      hard-to-understand areas.


Ivan-Mahda (approved):

packages/sdk/src/plt/v1/Token.ts
```diff
@@ -0,0 +1,254 @@
...
+export enum TokenErrorCode {
+    /**
+     * Error representing an attempt transfer funds to an account which is either not on the token allow list, or is on
+     * the token deny list
+     */
+    NOT_ALLOWED = 'NOT_ALLOWED',
+    /** Error representing an attempt to transfer tokens from an account that does not have enough tokens to cover the
+     * amount */
+    INSUFFICIENT_FUNDS = 'INSUFFICIENT_FUNDS',
+}
```
DOBEN, line 28: https://github.com/Concordium/concordium-grpc-api/pull/88/files#diff-a7337993ee30524fd2e3ee681d92ca3a483bb59389f312a741822917033175fdL864

Do we need the error case: 
- `non_existent_token_id`
- `invalid_account_reference`  when transferring to an account address that has not been created on chain.

packages/sdk/src/plt/v1/Token.ts
```diff
@@ -0,0 +1,254 @@
...
+     * Gets the name of the error, including its code.
+     * @returns {string} The name of the error.
+     */
+    public override get name(): string {
+        return `${this._name}.${this.code}`;
+    }
+}
+
+/**
+ * Error type indicating an unauthorized governance operation was attempted.
```
DOBEN, line 55:
```suggestion
 * Error type indicating an attempt transfer funds to an account which is either not on the token allow list, or is on the token deny list
```

packages/sdk/src/plt/v1/Token.ts
```diff
@@ -0,0 +1,254 @@
...
+    // Check that all receivers are NOT on the deny list (if present), or that they are included in the allow list (if present).
+    const receiverPromises = payloads.map((p) => token.grpc.getAccountInfo(p.recipient));
+    const receivers = await Promise.all(receiverPromises);
+    receivers.forEach((r) => {
+        const accToken =
+            r.accountTokens.find((t) => t.id.symbol === token.info.id.symbol) ??
+            bail(new NotAllowedError(r.accountAddress));
+        if (accToken.state.memberDenyList || accToken.state.memberAllowList === false) {
+            throw new NotAllowedError(r.accountAddress);
+        }
```
DOBEN, line 221: From my last information, the `denyList`/`allowList` will be checked on the `sender` address as well as the `receiver` address.

packages/sdk/src/extensions/cbor2.ts
```diff
@@ -1,2 +1,3 @@
-export * from '../types/cbor/encode.js';
-export * from '../types/cbor/decode.js';
+import { registerCBOREncoders } from '../types/cbor.js';
```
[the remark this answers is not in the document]
soerenbf: It's because the library it extends is called `cbor2` (which is the successor of the `cbor` library). So it's meant to communicate "if you're using `cbor2` in your project, you can add support for encoding the types in this library by using this "extension"

packages/sdk/src/types/cbor.ts
```diff
@@ -0,0 +1,114 @@
...
+    return encode(processedValue, { dcbor: true });
+}
+
+/**
+ * Registers all available CBOR decoders globally with the cbor2 library.
+ *
+ * This function currently registers the following decoders:
+ * - `AccountAddress` (tag 40307): For decoding Concordium account addresses
+ * - `TokenAmount` (tag 4): For decoding protocol-level token amounts as decimal fractions
+ * - `CborMemo` (tag 24): For decoding protocol-level token memos as cbor encoded data items
```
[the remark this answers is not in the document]
soerenbf: In the [cddl](https://github.com/Concordium/concordium-base/pull/607/files)

packages/sdk/src/plt/v1/Token.ts
soerenbf: Huh okay.. seems redundant in my head but will add it! Thanks

packages/sdk/src/plt/v1/Token.ts
soerenbf: `Token.fromId` gets the token info from chain and will fail if it does not exist. Account checking is up to the consumer I'd say. We don't do this for any transactions.

PR #424, opened 2025-05-12
Parse function for encoded v1 token module events

soerenbf:
## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/src/plt/v1/types.ts
+++ b/packages/sdk/src/plt/v1/types.ts
@@ -270,10 +270,10 @@ export function parseModuleEvent(event: EncodedModuleEvent): TokenModuleEvent {
 
     const decoded = Cbor.decode(event.details);
     if (typeof decoded !== 'object' || decoded === null) {
-        throw new Error(`Invalid event details: ${decoded}. Expected an object.`);
+        throw new Error(`Invalid event details: ${JSON.stringify(decoded)}. Expected an object.`);
     }
     if (!('target' in decoded) || !AccountAddress.instanceOf(decoded.target)) {
-        throw new Error(`Invalid event details: ${decoded}. Expected 'target' to be an AccountAddress`);
+        throw new Error(`Invalid event details: ${JSON.stringify(decoded)}. Expected 'target' to be an AccountAddress`);
     }
 
     const details: TokenListUpdateEventDetails = { target: { tag: 'account', address: decoded.target } };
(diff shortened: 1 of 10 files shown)


packages/sdk/src/plt/v1/types.ts
```diff
@@ -227,3 +195,90 @@ export type TokenInitializationParameters = {
...
+    /** The details of the event. */
+    details: T;
+};
+
+/**
+ * The structure of any list update event for a PLT V1 token.
+ */
+export type TokenListUpdateEventDetails = {
+    /** The target of the list update. */
+    target: TokenHolder;
```
soerenbf, line 211: This is different from how the target is decoded, which matches `TokenListUpdate`. The idea is to make it easier to discriminate between target variants when reading the data (which is the obvious use case for block events in general).

I was a little unsure if this was a good idea, or if we should just stick to describing whatever is decoded from the cbor encoded event details.

DOBEN: The current disadvantage seems to be that a new `tokenModule` in the future (with additional events) causes a breaking change with respect to the `parseModuleEvent` function because the function made further assumptions about the structure in `details`.

I suggest to have a generic function that just decodes the `details` from CBOR which could be used if people write an indexer and want to parse any `tokenModule` event (even future onces) without upgrading/breaking when new events are introduced.
`export function parseModuleEvent(event: EncodedModuleEvent)`

Suggestion would be to rename the current function to `parseTokenListUpdateEvent` that people can use if they are sure they are parsing a tx with that specific event.
`export function parseModuleEvent(event: EncodedModuleEvent)` to `parseTokenListUpdateEvent`

soerenbf: The `parseModuleEvent` is specifically scoped to parse events from the v1 token module, so a new token module comes along we'll add support for that with `v2/Token`.

The events can already be decoded without this helper function by simply doing `Cbor.decode(event)`. The `event` is of type `Cbor.Type`, which by convention means that the helper functions for the module is available on the `Cbor` module.

soerenbf: I'm deliberately trying to keep cbor parsing out of the _non-module_ code, as it's only the modules that concern themselves with cbor encoded details. We throw an error if the event cannot be parsed into a v1 token module event, in which case the user can chose to catch that and do manual parsing

PR #430, opened 2025-05-22
Add final `results` job to github CI pipeline

soerenbf:
To make it easier to define a rule in the settings of the repository that checks if the CI pipeline fails, we add a final job that always runs, but fails if previous jobs failed or were cancelled.

Taken from a discussion on the topic here: https://github.com/orgs/community/discussions/26822#discussioncomment-5122101

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -19,6 +19,7 @@ env:
 
 jobs:
   deps:
+    name: Dependencies
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-22.04
     steps:
@@ -48,6 +49,7 @@ jobs:
         run: yarn install --immutable
 
   build:
+    name: Build
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-22.04
     needs: deps
@@ -115,6 +117,7 @@ jobs:
             packages/react-components/dist
 
   tests:
+    name: Tests
     runs-on: ubuntu-22.04
     needs: build
     steps:
@@ -148,6 +151,7 @@ jobs:
         run: yarn test-ci
 
   typedoc:
+    name: Build docs
     runs-on: ubuntu-22.04
     needs: build
     steps:
@@ -178,6 +182,7 @@ jobs:
         run: yarn build:docs
 
   lint:
+    name: JS lint
     runs-on: ubuntu-22.04
     needs: build
     if: github.event.pull_request.draft == false
@@ -209,6 +214,7 @@ jobs:
         run: yarn lint:all && yarn fmt-check:all
 
   markdown-lint:
+    name: Markdown lint
     runs-on: ubuntu-22.04
     if: github.event.pull_request.draft == false
     needs: deps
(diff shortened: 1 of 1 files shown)


.github/workflows/pipeline.yml
```diff
@@ -335,3 +335,16 @@ jobs:
...
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    name: Final Results
+    needs: [tests, typedoc, lint, markdown-lint, markdown-linkcheck, rust_lint_fmt, rust_lint_clippy]
+    steps:
+      - run: exit 1
+        # see https://stackoverflow.com/a/67532120/4907315
+        if: >-
+          ${{
+               contains(needs.*.result, 'failure')
```
allanbrondum, line 348: Is `needs.build.result` the same as `needs.*.result`? Seems simpler then to be more explicit?

.github/workflows/pipeline.yml
```diff
@@ -335,3 +335,16 @@ jobs:
 
       - name: Run clippy
         run: cargo clippy --color=always --tests --benches -- -Dclippy::all
+  results:
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    name: Final Results
```
allanbrondum, line 341: Naming suggestion: 
```
action-result:
  ...
  name: Action Result
```

.github/workflows/pipeline.yml
allanbrondum, line 348: Is it better to formulate this as a "positive" condition. That `result` must be `'success'`?

.github/workflows/pipeline.yml
allanbrondum: Or `pipeline-result`

.github/workflows/pipeline.yml
soerenbf: I want to ensure that all previous builds succeed though, hence the wildcard. `needs.build.result` would only ensure that the `build` job (which is not a job dependency so it's invalid also) succeeds.

.github/workflows/pipeline.yml
soerenbf: Maybe `Workflow result`? I feel like `Action result` would refer to a single action in a job, while the intention is to communicate that the whole workflow executed successfully

.github/workflows/pipeline.yml
allanbrondum: Now I understand the notation :)

allanbrondum: But are there other results that should make us fail? (or if new are introduced). By having a positive list, we are sure that all job results leading to action/pipeline success are on the positive list where we have explicitly stated our opinion about the job result status.

allanbrondum: I think there is a time out result status also, though not sure. It is just easier to verify the correctness with a positive list

soerenbf: Yes I agree with you, but still there is no good way to do this other than expand a conditional with new branches when new jobs are added, as opposed to just adding a new entry in `needs`.

According to [the github docs on this matter](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#jobs-context) theres only `success`, `failure`, `skipped`  and `cancelled`.

The `needs` list of the `results` job _is_ the positive list.

allanbrondum (approved):

PR #432, opened 2025-05-22
Plt main

soerenbf:
## Purpose

Bring `plt` up to date with latest changes from `main`

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -19,6 +19,7 @@ env:
 
 jobs:
   deps:
+    name: Dependencies
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/pipeline.yml
```diff
@@ -335,3 +345,17 @@ jobs:
 
       - name: Run clippy
         run: cargo clippy --color=always --tests --benches -- -Dclippy::all
+  # The point of this job is to make it easy to add protection rules that require all status checks to pass.
+  results:
```
DOBEN, line 349:
```suggestion
        run: cargo clippy --color=always --tests --benches -- -Dclippy::all
        
  # The point of this job is to make it easy to add protection rules that require all status checks to pass.
  results:
```

DOBEN (approved):

PR #454, opened 2025-07-03
fix: webpack bundle bignumber.js alias

blorgon1:
## Purpose

Fixes an issue with the UMD bundle produced by Webpack. The exports for `bignumber.js` (a dependency of `json-bigint`) are not being handled correctly.

### Reproduction
To reproduce the issue on the current published bundle:
```
// require may require temporarily changing `"type": "commonjs"` in package.json to force node to import as a CJS module
const c = require('lib/min/concordium.node.min.js');
c.serializeTypeValue('a', Buffer.from([]), true);
```
Expected: (Not using real data for serialization, just getting past the use of `bignumber.js`)
```
Uncaught Error: Unable to serialize value due to: Parsing failed
```
Actual:
```
Uncaught TypeError: Right-hand side of 'instanceof' is not callable
```

## Changes

Add an alias in the webpack config to correct the exports of `bignumber.js`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/webpack.config.ts
+++ b/packages/sdk/webpack.config.ts
@@ -30,6 +30,10 @@ function configFor(target: 'web' | 'node' | 'react-native'): webpack.Configurati
                 '.js': ['.ts', '.js'],
             },
             extensions: ['.tsx', '.ts', '.js'],
+            alias: {
+                // Fix bignumber.js export issue - ensure it resolves to the BigNumber constructor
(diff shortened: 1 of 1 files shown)


soerenbf:
Thanks for contributing this fix @blorgon1, I was able to reproduce the issue and the issue is fixed withe your changes. Good find!

Will merge when CI passes, hope to have it released on a patch version soon after.

soerenbf:
@blorgon1 a new version of the sdk including this change has been released: https://www.npmjs.com/package/@concordium/web-sdk/v/9.2.0

PR #465, opened 2025-08-04
apply validation rules

Abutalif:
## Purpose

Applied the validation rules described [here](https://concordium.atlassian.net/wiki/spaces/EN/pages/1927348225/Protocol-level+Token+SDK+implemented+validation)

## Changes

- Removed authorization from the validation.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

- [x] I accept the above linked CLA.


packages/sdk/src/plt/Token.ts
```diff
@@ -510,7 +505,6 @@ export async function mint(
 
     if (validate) {
         token.moduleState.paused && bail(new PausedError(token.info.id));
-        validateGovernanceOperation(token, sender);
         amountsList.forEach((amount) => validateAmount(token, amount));
```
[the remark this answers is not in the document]
Abutalif: I think it is not possible at this moment, the validation of mint and burn are a bit different right now. I can factor it out into two different functions though:
```
validateMint(token: Token, amountsList) {
        token.moduleState.paused && bail(new PausedError(token.info.id));
        token.moduleState.mintable && bail(new NotMintableError(token.info.id));
        amountsList.forEach((amount) => validateAmount(token, amount));
}
```
and 
```
validateBurn(token: Token, amountsList) {
        token.moduleState.paused && bail(new PausedError(token.info.id));
        token.moduleState.burnable&& bail(new NotBurnableError(token.info.id));
        amountsList.forEach((amount) => validateAmount(token, amount));
}
````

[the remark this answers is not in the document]
Comment (soerenbf) [reply packages/sdk/src/plt/Token.ts]: Makes sense, I think we should do that to make the functions available separately to users as well

File: packages/sdk/src/plt/Token.ts [hunk 2]
```diff
@@ -180,6 +172,54 @@ export class PausedError extends TokenError {
...
+    public readonly code = TokenErrorCode.NO_LIST;
+
+    /**
+     * Constructs a new NoListError.
+     *
+     * @param {TokenId.Type} tokenId - The ID of the token.
+     */
+    constructor(public readonly tokenId: TokenId.Type) {
+        super(`Token ${tokenId} does not have allow/deny list.`)
+    }
```
soerenbf, line 220:
```suggestion
    constructor(public readonly tokenId: TokenId.Type, list: string) {
        super(`Token ${tokenId} does not have ${list} list.`)
    }
```

packages/sdk/src/plt/Token.ts
```diff
@@ -180,6 +172,54 @@ export class PausedError extends TokenError {
...
+}
```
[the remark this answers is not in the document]
Abutalif: The Rust counterpart wiil have two different error types for this: `NoAllowList`, `NoDenyList`.
Should I do the same here, to ensure allignment between two SDKs, or stick to the suggestion above. It seems like a more idiomatic approach for JS.

packages/sdk/src/plt/Token.ts
```diff
@@ -198,6 +279,14 @@ class Token {
...
         this.moduleState = Cbor.decode(info.state.moduleState, 'TokenModuleState');
     }
+
+    /**
+     * Returns a new Token instance with updated info and parsed module state.
+     */
+    public async updateToken(): Promise<Token> {
+        const updatedInfo = await this.grpc.getTokenInfo(this.info.id);
+        return new Token(this.grpc, updatedInfo);
+    }
```
soerenbf, line 289: I would prefer if we mutated the existing token here, I don't see any reason for the user to keep outdated token info around. Can still a reference to the updated token for convenience 🙂 

The way to do this in js/ts would be like this:

```ts
class Test {
  private another: number;

  constructor(private prop: string) {
    this.another = Number(prop);
  }

  public get prop(): string {  return this.prop; }
  public get another(): number {  return this.another; }
}
```

Please remember to update the docstring for `updateToken` and mention that it mutates the internal state of the token.

packages/sdk/src/plt/Token.ts
```diff
@@ -375,30 +465,111 @@ export async function validateTransfer(
...
+
+    const { decimals } = updatedToken.info.state;
+    const burnableAmount = (await balanceOf(updatedToken, sender)) ?? TokenAmount.zero(decimals);
+    const payloadTotal = amountsList.reduce(
+        (acc, amount) => acc.add(TokenAmount.toDecimal(amount)),
+        TokenAmount.toDecimal(TokenAmount.zero(decimals))
+    );
+    if (TokenAmount.toDecimal(burnableAmount).lt(payloadTotal)) {
+        throw new InsufficientSupplyError(sender, TokenAmount.fromDecimal(payloadTotal, decimals));
+    }
```
soerenbf, line 541: Nice addition! Great job 😄

packages/sdk/test/ci/plt/Token.test.ts
soerenbf, line 1: I would love it if we could add some tests for the added validation functions 🙏 Should be enough around in the other tests to see how to do mocking of the grpc api, which is the primary challenge I think.

soerenbf (commented):
Loving the amount of detail, great job 🎉 I have one final round of changes I'd like to see before we merge.

packages/sdk/CHANGELOG.md
```diff
@@ -55,12 +55,18 @@
 
 - Enable `denyList`/`allowList` validation on plt token transfers.
 - Client side validation in the PLT `Token` client is now disabled by default and has to be enabled explicitly if wanted
+- Remove authorization validation for PLT `Token` client.
```
DOBEN, line 58: Should be under the `## 10.0.0-alpha.? (Unreleased)` section.

packages/sdk/src/plt/Token.ts
```diff
@@ -399,6 +501,101 @@ export async function validateTransfer(
...
+    let senderBalance: BalanceOfResponse;
+    let senderAdderss: AccountAddress.Type;
+
+    if (AccountAddress.instanceOf(sender)) {
+        senderAdderss = sender;
+        senderBalance = await balanceOf(token, sender);
+    } else {
+        senderAdderss = sender.accountAddress;
+        senderBalance = balanceOf(token, sender);
+    }
```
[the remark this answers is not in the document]
Abutalif: This version is more concise, but defaults to using the async version of the `balanceOf`. I think it is unneccessary, as users may already have `AccountInfo`, so no need to fetch it again from the node.

PR #476, opened 2025-08-19
Bugfix/support token creation

soerenbf:
Adds support for transaction summary type `tokenCreation`, which is currently missing and causes a runtime error.

## Checklist

      hard-to-understand areas.


packages/sdk/CHANGELOG.md
```diff
@@ -2,9 +2,14 @@
 
 ## Unreleased
 
-## 10.0.0
+## 10.0.1
+
+### Fixed
 
-Adds support for integrating with Concordium nodes running version 9.
```
DOBEN, line 7: I think we should keep that info at the major version `10.0.0` .

PR #491, opened 2025-09-03
Wrapped NodeInfoConsensusStatus and NodeCatchupStatus with Upward

rbaconcordium:
Jira-Id: COR-1841

## Purpose

Ensure new variants in gRPC API for NodeCatchupStatus and NodeInfoConsensusStatus are forward-compatible.

## Changes

Wrapped using Upward

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/src/grpc/translation.ts
+++ b/packages/sdk/src/grpc/translation.ts
@@ -2460,7 +2460,7 @@ export function nodeInfo(nodeInfo: GRPC.NodeInfo): SDK.NodeInfo {
     };
 }
 
-function trCatchupStatus(catchupStatus: GRPC.PeersInfo_Peer_CatchupStatus): SDK.NodeCatchupStatus {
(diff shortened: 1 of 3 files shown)


td202 (approved):

packages/sdk/src/grpc/translation.ts
```diff
@@ -2469,6 +2469,8 @@ function trCatchupStatus(catchupStatus: GRPC.PeersInfo_Peer_CatchupStatus): SDK.
             return SDK.NodeCatchupStatus.Pending;
         case CatchupStatus.UPTODATE:
             return SDK.NodeCatchupStatus.UpToDate;
+        default:
+            return null;
```
soerenbf, line 2473: A bit late to the party here, but instead we should use 
```suggestion
        case undefined:
            return null;
```

This will help us notice new variants when updating the SDK with a new reference to the protobuf definitions

PR #516, opened 2025-09-30
Feature/cor 1171 upgrade wasm rust step2

rbaconcordium:
## Purpose

Update wasm-pack and rust version above 1.74

## Changes

Have to modify a few json files to get rust-bindings project to build. There is also a need to introduce a new separate step to compile the ts files in the scripts directory.

## Checklist

      hard-to-understand areas.


packages/rust-bindings/scripts/build-react-native-wallet.ts
```diff
@@ -1,9 +1,15 @@
...
+//import { convertWasmToJs, copyToFolder } from '../dist/scripts/build-scripts.js';
+import { copyToFolder } from '../dist/scripts/build-scripts.js';
+
 
 const walletBundlerPath = path.join(__dirname, '../lib/wallet/bundler');
 const walletOutPath = path.join(__dirname, '../lib/wallet/react-native');
 
 copyToFolder(walletBundlerPath, walletOutPath);
-convertWasmToJs(walletOutPath);
+//convertWasmToJs(walletOutPath);
```
soerenbf, line 15: We cannot just remove this, as this step is needed specifically for the bindings to be compatible with react-native.

packages/rust-bindings/scripts/build-scripts.ts
```diff
@@ -41,3 +42,4 @@ export const convertWasmToJs = async (outPath: string) => {
             fs.writeFileSync(file, content, 'utf-8');
         });
 };
+*/
```
soerenbf, line 45: What's the intention here? This is a required part of the compilation for react-native compatibility

packages/rust-bindings/package.json
```diff
@@ -82,12 +83,14 @@
         "./lib/**/index*"
     ],
     "scripts": {
+        "compile-build-script": "tsc -p tsconfig.scripts-build-script.json",
+        "compile-scripts": "yarn compile-build-script && tsc -p tsconfig.scripts.json",
```
soerenbf, line 87: Why are these build steps required all of a sudden? We even still use the `.ts` versions of the build scripts related to the react-native target.

If you can clean it up, please also remember to clean up the added tsconfigs, which in my view should not be needed.

packages/rust-bindings/tsconfig.build.json
```diff
@@ -1,6 +1,6 @@
 {
     "extends": "./tsconfig.json",
-    "include": ["./ts-src/**/*"],
+    "include": ["./ts-src/**/*", "./wasm.d.ts"],
```
soerenbf, line 3: Please also add this to the base `tsconfig.json`

packages/rust-bindings/ts-src/dapp.ts
```diff
@@ -1,12 +1,12 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import { Buffer } from 'buffer/index.js';
 
-import { initSync } from '../lib/dapp/web/esm';
+import { initSync } from '../lib/dapp/web/esm/index.js';
 import wasmBase64 from '../lib/dapp/web/esm/index_bg.wasm';
 
 // Expected to resolve to base64 encoded bytes of wasm module
 
 const bytes = Buffer.from(wasmBase64 as unknown as string, 'base64');
```
soerenbf, line 9: Is it possible to remove the `as unknown as string` force cast now with the added `wasm.d.ts` declaration file?

packages/rust-bindings/wasm.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.wasm" {
```
soerenbf, line 1: I think it would be good to add an explanation to this:
```suggestion
/**
 * Represents the format expected from importing `.wasm` files the have been pre-processed
 * to base64 strings as part of a build job.
 */
declare module "*.wasm" {
```

packages/rust-bindings/package.json
rbaconcordium: the reason for this from what i can understand so far, is the import line needing the explicit extension, in this case we need .js
With the upgrade, we seem to be getting a more strict rule where the extension needs to be specified.
i cannot put .ts, as that seems to require some kind config allowTs... = true with noEmit being true.

So, I think, I went around this way, to explicitly generate .js from the ts in a separate directory by this step.

PR #517, opened 2025-09-30
Feature/cor 1171 upgrade wasm rust step3

rbaconcordium:
## Purpose

_Describe the purpose of the pull request, link to issue describing the problem, etc.

## Changes

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/packages/sdk/src/cis0.ts
+++ b/packages/sdk/src/cis0.ts
@@ -139,7 +139,9 @@ export async function cis0Supports(
         return undefined;
     }
 
-    const parameter = Parameter.fromBuffer(makeDynamicFunction(serializeSupportIdentifiers)(standardIds));
+    const parameter = Parameter.fromBuffer(
+        (makeDynamicFunction(serializeSupportIdentifiers)(standardIds) as any as Buffer).buffer as ArrayBuffer
+    );
 
     const response = await grpcClient.invokeContract(
         {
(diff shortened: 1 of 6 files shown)


.github/workflows/pipeline.yml
```diff
@@ -14,7 +14,7 @@ on:
 env:
   DUMMY: 3 # For cache busting.
   NODE_VERSION: 18.16.0
-  RUST_VERSION: 1.73
+  RUST_VERSION: 1.74
```
soerenbf, line 17: Needs to be at least
```suggestion
  RUST_VERSION: 1.78
```
to be compatible with lock version 4 (from `cargo.lock`). I think we should go with the highest version you can get it to build with though, i.e. if it works with version 1.85 on your machine, let's specify that here as well.

packages/sdk/src/accountTransactions.ts
```diff
@@ -108,7 +108,10 @@ export class SimpleTransferHandler
     }
 
     deserialize(serializedPayload: Cursor): SimpleTransferPayload {
-        const toAddress = AccountAddress.fromBuffer(Buffer.from(serializedPayload.read(32)));
+        // 1. Correct the INPUT type conflict using the triple assertion
+        const inputBuffer = Buffer.from(serializedPayload.read(32)).buffer as any as ArrayBuffer;
+
+        const toAddress = AccountAddress.fromBuffer(inputBuffer);
```
soerenbf, line 114: I'm a hesitant to approve of this without knowing why this happens, i.e. why has it been possible up to this point and now all of a sudden we need to do force casts to make it compile 🤔

I think we need to look into whether the `buffer` dependency version changed and why that would be incompatible with the current signature. We generally use `ArrayBuffer` as the type to represent memory buffers (usually a byte array) for input to functions everywhere. It would not be nice to have to cast like this for all uses of the functions in question as that makes us lose compile time errors for all of this.

soerenbf:
Done in #519

PR #529, opened 2025-10-07
WS upgrade to 8.17.1 due to vulnerability alert

rbaconcordium:
## Purpose

WS upgrade to 8.17.1 due to vulnerability alert

## Changes

upgraded ws version to 8.17.1 due to vulnerability alert
had to add a new line in resolutions of package.json
ran yarn up ws and parents, version stayed as before
ran yarn npm audit, the vulnerability alert is gone

## Checklist

      hard-to-understand areas.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
     },
     "resolutions": {
         "@walletconnect/types": "2.13.0",
-        "@protobuf-ts/runtime-rpc": "2.9.1"
+        "@protobuf-ts/runtime-rpc": "2.9.1",
+        "path-to-regexp": "0.1.12"
     },
     "scripts": {
         "test-ci": "yarn workspace @concordium/web-sdk run test-ci && yarn workspaces foreach --no-private --exclude @concordium/web-sdk run test",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "resolutions": {
         "@walletconnect/types": "2.13.0",
         "@protobuf-ts/runtime-rpc": "2.9.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -35,7 +35,10 @@
     },
     "resolutions": {
         "@walletconnect/types": "2.13.0",
-        "@protobuf-ts/runtime-rpc": "2.9.1"
+        "@protobuf-ts/runtime-rpc": "2.9.1",
+        "path-to-regexp": "0.1.12",
+        "body-parser": "1.20.3",
+        "ws": "8.17.1"
```
soerenbf, line 39: try

```sh
git reset main --hard
yarn up ws -R
```

I could get it to upgrade `ws@8` to `ws@8.13.3`

soerenbf: Also, if you we were to specify a resolution for `ws`, this would not be the ideal way to do it.

```jsonc
"resolutions" : {
  "ws": "8.17.1" // every version of ws will be resolved to this version, even previous and future major versions
}
```

the `ws` dependency is used in many different versions (if you look at the output of `yarn why ws`), so forcing all versions onto major version 8 will almost certainly break things.

So what you'd need to do is scope the force resolve to the dependencies where this is actually an issue (documented here: https://yarnpkg.com/configuration/manifest#resolutions)

soerenbf (changes requested):

yarn.lock
soerenbf, line 1: what happened here? 14.5k lines changed? 🙈

PR #530, opened 2025-10-07
rollup upgrade to newer version due to vulnerability alert

rbaconcordium:
## Purpose

Upgraded rollup version to 3.29.5 due to vulnerability alert for the older version that was used in the transitive dependency

## Changes

Added a new line in resolutions in package.json to set to 3.29.5.
ran yarn npm audit to confirm the vulnerability is no longer alerted.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
     },
     "resolutions": {
         "@walletconnect/types": "2.13.0",
-        "@protobuf-ts/runtime-rpc": "2.9.1"
+        "@protobuf-ts/runtime-rpc": "2.9.1",
+        "path-to-regexp": "0.1.12"
     },
     "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,7 +38,8 @@
         "@protobuf-ts/runtime-rpc": "2.9.1",
         "path-to-regexp": "0.1.12",
         "body-parser": "1.20.3",
-        "ws": "8.17.1"
+        "ws": "8.17.1",
+        "rollup": "3.29.5"
```
soerenbf, line 42: please do `yarn up rollup -R` instead
```suggestion
```

soerenbf (changes requested):

package.json
rbaconcordium: i will do another clean PR for this, as the previous PR with ws I restarted, just to be safe

PR #588, opened 2025-11-07
Remove maxContractExecutionEnergy from UpdateContractPayload.

rbaconcordium:
## Purpose

Remove maxContractExecutionEnergy from UpdateContractPayload. For contracts, this energy field is not derived from the payload. 
It should be calculated separately and pass in manually during serialization.

## Changes

UpdateContractPayload type no longer has the maxContractExecutionEnergy. Serialize functions for transactions now have overloaded signatures to deal with InitContract and UpdateContract and to accept the energy amount separately instead of coming from payload. The energy field in InitContract will be dealt in a separate PR, for information.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/test/ci/accountTransactions.test.ts
+++ b/packages/sdk/test/ci/accountTransactions.test.ts
@@ -37,7 +37,7 @@ import {
     UpdateCredentialsPayload,
     getAccountTransactionHandler,
     getAccountTransactionSignDigest,
-    serializeAccountTransactionPayload
+    serializeAccountTransactionPayload,
 } from '../../src/index.js';
 import { AccountAddress, TransactionExpiry } from '../../src/pub/types.js';
 
--- a/packages/sdk/test/ci/deserialization.test.ts
+++ b/packages/sdk/test/ci/deserialization.test.ts
@@ -142,7 +142,12 @@ test('test deserialize UpdateContract ', () => {
         message: Parameter.fromHexString('0a'),
     };
 
-    const transaction = prepareTransaction(AccountTransactionType.Update, deserializePayload, undefined, Energy.create(3000));
+    const transaction = prepareTransaction(
+        AccountTransactionType.Update,
+        deserializePayload,
+        undefined,
+        Energy.create(3000)
+    );
 
     deserializeAccountTransactionBase(transaction);
 });
@@ -232,7 +237,7 @@ function prepareTransaction<T extends AccountTransactionType, P extends AccountT
     transactionType: T,
     payload: P,
     inputExpiry?: TransactionExpiry.Type,
-    givenEnergy?: Energy.Type,
+    givenEnergy?: Energy.Type
 ) {
     const handler = getAccountTransactionHandler(transactionType);
 


packages/sdk/src/serialization.ts
```diff
@@ -144,9 +196,27 @@ export function serializeAccountTransactionPayload(accountTransaction: Omit<Acco
...
+            signatures,
+            givenEnergy
+        );
+    } else {
+        serializedAccountTransaction = serializeAccountTransaction(
+            accountTransaction as AccountTransaction<OtherType, OtherPayload>,
+            signatures
+        );
+    }
+
```
soerenbf:
```suggestion
    const serializedAccountTransaction = serializeAccountTransaction(accountTransaction, signatures);
```

Does this not still work? Otherwise it might make sense to include a completely dynamic (internal, i.e. _not_ exported) overload of it:

```
function serializeAccountTransaction(
    accountTransaction: AccountTransaction,
    signatures: AccountTransactionSignature,
    givenEnergy?: Energy.Type
): Buffer;
```

soerenbf (approved):
Nice job, a few comments:

packages/sdk/src/serialization.ts
rbaconcordium: if we didn't use the IF ELSE, there is a compiler error stating "Argument of type 'AccountTransaction<AccountTransactionType, AccountTransactionPayload>' is not assignable to parameter of type 'AccountTransaction<AccountTransactionType.DeployModule | AccountTransactionType.Transfer | AccountTransactionType.AddBaker | AccountTransactionType.RemoveBaker | ... 15 more ... | AccountTransactionType.TokenUpdate, SimpleTransferPayload | ... 6 more ... | TokenUpdatePayload>'.
  Type 'AccountTransactionType' is not assignable to ty...."

packages/sdk/src/serialization.ts
```diff
@@ -181,9 +271,28 @@ export function getAccountTransactionSignDigest(accountTransaction: AccountTrans
...
+            accountTransaction as AccountTransaction<InitUpdateType, InitUpdatePayload>,
+            signatures,
+            givenEnergy
+        );
+    } else {
+        serializedAccountTransaction = serializeAccountTransaction(
+            accountTransaction as AccountTransaction<OtherType, OtherPayload>,
+            signatures
+        );
+    }
```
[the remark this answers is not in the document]
rbaconcordium: yup same compiler issue here as previous...

packages/sdk/src/accountTransactions.ts
```diff
@@ -51,6 +52,8 @@ export interface AccountTransactionHandler<
     PayloadType extends AccountTransactionPayload = AccountTransactionPayload,
     JSONType = PayloadType,
 > {
+    create(metadata: TransactionMetadata, payload: PayloadType, givenEnergy?: Energy.Type): AccountTransaction;
```
td202: Add docs. Also, maybe this should be formatted more like the other operations:

```suggestion
    create: (metadata: TransactionMetadata, payload: PayloadType, givenEnergy?: Energy.Type) => AccountTransaction;
```

packages/sdk/src/types.ts
```diff
@@ -1425,6 +1421,12 @@ export interface AccountTransactionHeader {
 
     /** expiration of the transaction */
     expiry: TransactionExpiry.Type;
+
+    /** energy amount */
+    energyAmount: Energy.Type;
```
td202: Consider renaming to `baseEnergyAmount` and document that this excludes transaction size and signature costs. Or possibly not "base" because that is also used to refer to the transaction size + signatures part of the cost.

PR #591, opened 2025-11-07
create audit record checked

soerenbf:
## Purpose

Adds a function for creating an audit record based on a verification of the presentation both cryptographically (from `concordium-base`) and in the context of the origin verification request.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/src/wasm/VerifiablePresentationV1/audit-record.ts
+++ b/packages/sdk/src/wasm/VerifiablePresentationV1/audit-record.ts
@@ -192,14 +192,23 @@ function verifyAtomicStatements<A>(a: AtomicStatementV2<A>[], b: AtomicStatement
                 if (bs.type !== 'AttributeInSet') throw atomicError;
                 if (as.set.length !== bs.set.length) throw atomicError;
                 // For the sets, not all implementations used lists where the order items are added
-                // is enforced (i.e. some use Sets). We expect no duplicates in the sets.
-                if (!as.set.every((asv) => bs.set.some((bsv) => attributeTypeEquals(asv, bsv)))) throw atomicError;
+                // is enforced (i.e. some use Sets).
+                //
+                // This is O(2*n^2), but we don't expect a lot of elements to compare.
+                const setEqual =
+                    as.set.every((asv) => bs.set.some((bsv) => attributeTypeEquals(asv, bsv))) &&
+                    bs.set.every((bsv) => as.set.some((asv) => attributeTypeEquals(asv, bsv)));
+                if (!setEqual) throw atomicError;
                 break;
             }
             case 'AttributeNotInSet': {
                 if (bs.type !== 'AttributeNotInSet') throw atomicError;
                 if (as.set.length !== bs.set.length) throw atomicError;
-                if (!as.set.every((asv) => bs.set.some((bsv) => attributeTypeEquals(asv, bsv)))) throw atomicError;
+
+                const setEqual =
+                    as.set.every((asv) => bs.set.some((bsv) => attributeTypeEquals(asv, bsv))) &&
+                    bs.set.every((bsv) => as.set.some((asv) => attributeTypeEquals(asv, bsv)));
+                if (!setEqual) throw atomicError;
                 break;
             }
             case 'RevealAttribute': {


examples/nodejs/proofs/verifiable-presentation-id-cred.ts
```diff
@@ -189,20 +192,23 @@ const presentation = await VerifiablePresentationV1.createFromAnchor(
 // simulate sending a response from the application to the client requesting the presentation
 const presentationJson = JSONBig.stringify(presentation);
 
-console.log('PRESENTATION:\n', JSONBig.stringify(presentation, null, 2), '\n');
-
 // simulate receiving presentation to be verified
 const presentationParsed = VerifiablePresentationV1.fromJSON(JSONBig.parse(presentationJson));
 
-if (!(await VerifiablePresentationV1.verifyWithNode(presentationParsed, grpc, network)))
```
DOBEN: This is temporary removed, isn't it?

packages/sdk/src/wasm/VerifiablePresentationV1/audit-record.ts
```diff
@@ -101,6 +114,212 @@ export function create(
...
+        'Mismatch found when comparing atomic statements for verification/presentation request.'
+    );
+
+    a.forEach((as, i) => {
+        const bs = b[i];
+        switch (as.type) {
+            case 'AttributeInSet': {
+                if (bs.type !== 'AttributeInSet') throw atomicError;
+                if (as.set.length !== bs.set.length) throw atomicError;
+                if (!as.set.every((asv) => bs.set.includes(asv))) throw atomicError;
```
DOBEN: Is it actually enforced that `set` is a mathematical `set` (no element can appear more then once).

They `set` field is an array, so it could include duplicates.

E.g. the following would parse in your example as equal, isn't it?
as = [a, a]
bs = [a, b]

packages/sdk/src/wasm/VerifiablePresentationV1/audit-record.ts
```diff
@@ -101,6 +114,212 @@ export function create(
...
+                break;
+            }
+            case 'AttributeNotInSet': {
+                if (bs.type !== 'AttributeNotInSet') throw atomicError;
```
DOBEN: What about the attributeTag. Should be compared as well. Otherwise one statement is about citizenship but the other statement is about nationality

packages/sdk/src/wasm/VerifiablePresentationV1/audit-record.ts
```diff
@@ -101,6 +114,212 @@ export function create(
...
+                if (as.set.length !== bs.set.length) throw atomicError;
+                if (!as.set.every((asv) => bs.set.includes(asv))) throw atomicError;
```
DOBEN: same about the set comparision

packages/sdk/src/wasm/VerifiablePresentationV1/audit-record.ts
```diff
@@ -101,6 +114,212 @@ export function create(
...
+                break;
+            }
+            case 'RevealAttribute': {
+                if (bs.type !== 'RevealAttribute') throw atomicError;
```
DOBEN: attributeTag is missing.

packages/sdk/src/wasm/VerifiablePresentationV1/audit-record.ts
```diff
@@ -101,6 +114,212 @@ export function create(
...
+        return { type: 'failed', error: e as Error };
+    }
+
+    // 2. verify cryptographic integrity of presentation
+    const verification = await VerifiablePresentationV1.verifyWithNode(presentation, grpc, network, blockHash);
+    if (verification.type === 'failed') return verification;
+
+    // 3. Checck that none of the credentials have expired
+    verification.result.credentialsStatus.every((cs) => cs === CredentialStatus.Active) ||
+        bail('One or more credentials included in the presentation has expired.');
```
DOBEN:
```suggestion
        bail('One or more credentials included in the presentation is not active.');
```

Can also be `non-active/revoked`?

packages/sdk/src/wasm/VerifiablePresentationV1/audit-record.ts
soerenbf: had a look and seems you're right so will revise. Also this will not work when `AttributeType` variant is `TimestampAttribute` as that would check object eq so need to rework anyway.

packages/sdk/src/wasm/VerifiablePresentationV1/audit-record.ts
```diff
@@ -101,6 +114,212 @@ export function create(
...
+
+    const atomicError = new Error(
+        'Mismatch found when comparing atomic statements for verification/presentation request.'
+    );
+
+    a.forEach((as, i) => {
+        const bs = b[i];
+        switch (as.type) {
+            case 'AttributeInSet': {
+                if (bs.type !== 'AttributeInSet') throw atomicError;
```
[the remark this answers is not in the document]
soerenbf: good catch

packages/sdk/src/wasm/VerifiablePresentationV1/audit-record.ts
```diff
@@ -101,6 +114,212 @@ export function create(
...
+                if (as.upper !== bs.upper) throw atomicError;
+                break;
+            }
+        }
+    });
+}
+
+async function verifyAccountClaims(
+    claims: AccountClaims,
+    statement: VerificationRequestV1.Statement,
```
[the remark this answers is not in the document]
soerenbf: Do you mean going through the atomic statements to see if they match? This is checked in `verififyClaims`.

```
    if (statement.statement.length !== claims.statement.length)
        throw new Error('Mismatch in number of atomic statements for statement/claim');

    verifyAtomicStatements(statement.statement, claims.statement);

    if (isAccountClaims(claims)) {
        await verifyAccountClaims(claims, statement, grpc);
    } else {
        verifyIdentityClaims(claims, statement);
    }
```

packages/sdk/src/wasm/VerifiablePresentationV1/audit-record.ts
soerenbf: yeah might as well.. cannot ensure the list order is enforced as the rust code deserializes into a btree set, but will do the alternative implementation.

PR #594, opened 2025-11-10
Update base ref to sponsored transactions branch

soerenbf:
Ref SPO-1

## Purpose

The initial update of the base reference to include the sponsored transactions GRPC API changes

## Checklist

      hard-to-understand areas.


packages/sdk/src/plt/Cbor.ts
soerenbf: The type was renamed in the protobuf files, which is why this change was needed.

PR #602, opened 2025-11-14
Transaction revision

soerenbf:
Closes SPO-44

## Purpose

In preparation of adding support for the extended transaction format, this adds a new `Transaction` and `Payload` API. which aims to describe on-chain representation of account transactions v0 (+ v1 in future) and provide functionality for building these.

The new API is supposed to both align with transaction APIs in other libraries and help users build complex transactions, i.e. multi-sig, sponsored, future extensions, and any **valid** combination of these with minimal fees.

This reverts _some_ of the breaking changes on the target branch `release/sdk/12` and implements them in these new APIs instead.

## Changes

- Deprecated `AccountTransactionHandler` and its implementations.
- Added support for submitting the new transaction formats in the GRPC client.
- Add `AccountTransactionInput` which replaces `AccountTransactionPayload` as it was previously defined. This will ease migration for our wallets depending on the associated variants.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/test/ci/transactions/AccountTransactionV0.test.ts
+++ b/packages/sdk/test/ci/transactions/AccountTransactionV0.test.ts
@@ -65,7 +65,7 @@ describe('AccountTransactionV0', () => {
             };
             expect(json.header).toEqual(expectedHeader);
 
-            const roundtrip = AccountTransactionV0.fromJSON(JSONBig.parse(JSONBig.stringify(json)), 'signed');
+            const roundtrip = AccountTransactionV0.fromJSON(JSONBig.parse(JSONBig.stringify(json)));
             expect(roundtrip).toEqual(transaction);
         });
     });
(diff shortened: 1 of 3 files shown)


packages/sdk/src/transactions/AccountTransactionV0.ts
```diff
@@ -0,0 +1,335 @@
...
+/**
+ * Converts a intermediary JSON serializable representation created with {@linkcode toJSON} back to a
+ * version 0 account transaction.
+ *
+ * @param json the JSON to convert
+ * @param [as] the transaction variant to parse. Defaults to parsing {@linkcode AccountTransactionV0}.
+ * @returns the transaction
+ */
+export function fromJSON(json: JSON, as?: 'signed'): AccountTransactionV0;
+export function fromJSON(json: UnsignedJSON, as: 'unsigned'): Unsigned;
```
limemloh: This seems a bit off to me, overloading the parameters and return type like this...
Does it have to be one function or could it be split into two separate functions?
Similar comment to the other `toJSON`/`fromJSON` variants

PR #659, opened 2026-03-02
RBAC work, adding updateMetadata and updateAdminRoles operations

rbaconcordium:
## Purpose
To introduce two operations, to update metadata and to update admin roles.

## Changes
Added two functions, updateMetadata and updateAdminRoles.
Added TokenOperationType.UpdateMetadata and TokenOperationType.AssignAdminRoles, RevokeAdminRoles.
Added enum TokenAdminRole
Added type TokenUpdateAdminRoleDetails
Added operation types TokenUpdateAdminRoleOperation, TokenAssignAdminRolesOperation, TokenRevokeAdminRolesOperation and TokenUpdateMetadataOperation

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


packages/sdk/src/plt/Token.ts
```diff
@@ -902,6 +907,78 @@ export async function unpause(
...
+): Promise<TransactionHash.Type> {
+    switch (type) {
+        case TokenOperationType.AssignAdminRoles:
+            return assignAdminRoles(token, updateAdminRoleDetails, sender, signer, metadata);
+        case TokenOperationType.RevokeAdminRoles:
+            return revokeAdminRoles(token, updateAdminRoleDetails, sender, signer, metadata);
+        default:
+            throw new Error(`Unsupported operation type: ${type}`);
+    }
+}
```
soerenbf: I don't think we need this. The type in the cddl is mostly there to collect similar operations under a single name but it's not useful I think.

packages/sdk/src/plt/TokenOperation.ts
```diff
@@ -14,6 +14,20 @@ export enum TokenOperationType {
...
+}
+
+export enum TokenAdminRole {
+    UpdateAdminRoles = 'updateAdminRoles',
+    Mint = 'mint',
+    Burn = 'burn',
+    UpdateAllowList = 'allowList',
+    UpdateDenyList = 'denyList',
+    Pause = 'pause',
+    Unpause = 'unpause',
```
soerenbf:
```suggestion
```

There's no "unpause" role, the "pause" role will be used for this purpose as well 🙂

soerenbf (commented):
Looks good in general. Here are a few things I think would be good to add/change:

- Change the target branch to a feature branch that collects functionality for the feature (as discussed)
- Add unit tests for the CBOR encoding of the operations added
- Update the `CHANGELOG.md`

packages/sdk/src/plt/Token.ts
rbaconcordium: removing this function and exporting the individual functions instead.

packages/sdk/src/plt/TokenOperation.ts
rbaconcordium: removing

rbaconcordium:
> Looks good in general. Here are a few things I think would be good to add/change:
> 
> * Change the target branch to a feature branch that collects functionality for the feature (as discussed)
> * Add unit tests for the CBOR encoding of the operations added
> * Update the `CHANGELOG.md`

I have changed as per above.

packages/sdk/src/plt/Token.ts
```diff
@@ -902,6 +907,69 @@ export async function unpause(
...
+ * @param {Token} token - The token to update.
+ * @param {MetadataUrl} metadataUrl - The new metadata URL.
+ * @param {AccountAddress.Type} sender - The account address of the sender.
+ * @param {AccountSigner} signer - The signer responsible for signing the transaction.
+ * @param {TokenUpdateMetadata} [metadata={ expiry: TransactionExpiry.futureMinutes(5) }] - The metadata for the token update.
+ * @returns A promise that resolves to the transaction hash.
+ */
+export async function updateMetadata(
+    token: Token,
+    metadataUrl: MetadataUrl,
```
td202: Might it make sense to allow calling this function with a string (representing the raw URL) as a convenience?

packages/sdk/src/plt/TokenOperation.ts
```diff
@@ -31,6 +44,24 @@ export type TokenTransfer = {
...
 };
 
+/**
+ * URL identifies a metadata together with an optional sha256 checksum of the contents of the metadata
+ */
+export type MetadataUrl = {
+    /** The URL of the metadata. */
+    url: string;
+    /** An optional SHA256 checksum of the metadata contents. */
+    checksum?: string;
```
squirer: we have an additional fields which exists in base here: https://github.com/Concordium/concordium-base/blob/main/rust-src/concordium_base/src/protocol_level_tokens/token_metadata_url.rs#L47

its probably worth checking if this also needs to be added in the SDK here - i would imagine everything should be 1:1

td202 (approved):

packages/sdk/src/plt/TokenOperation.ts
rbaconcordium: removing this, there is an existing type already written for TokenMetadataUrl

packages/sdk/src/plt/Token.ts
rbaconcordium: changing to use an existing TokenMetadataUrl type, that already has a function to create this structure via create function, e.g. TokenMetadataUrl.create
        let tokenMetadataUrl = TokenMetadataUrl.create(url, checksum, additional);

squirer (approved):

packages/sdk/src/plt/Token.ts
```diff
@@ -902,6 +907,69 @@ export async function unpause(
...
+ * Updates the metadata URL of a token.
+ *
+ * @param {Token} token - The token to update.
+ * @param {TokenMetadataUrl.Type} metadataUrl - The new metadata URL.
+ * @param {AccountAddress.Type} sender - The account address of the sender.
+ * @param {AccountSigner} signer - The signer responsible for signing the transaction.
+ * @param {TokenUpdateMetadata} [metadata={ expiry: TransactionExpiry.futureMinutes(5) }] - The metadata for the token update.
+ * @returns A promise that resolves to the transaction hash.
+ */
+export async function updateMetadata(
```
limemloh: **Nitpicking (non-blocking):** No need for the `async` keyword here, since you never use `await`.

It is better to not have it, since it actually impose some runtime overhead.

Same for the functions below

packages/sdk/src/plt/TokenOperation.ts
```diff
@@ -31,6 +44,14 @@ export type TokenTransfer = {
     memo?: Memo;
 };
 
+/**
+ * The details required to update an admin role for a token.
+ */
+export type TokenUpdateAdminRolesDetails = {
+    roles: TokenAdminRole[];
+    account: CborAccountAddress.Type;
+};
```
limemloh, line 57: **Minor thing (non-blocking):** Please add documentation for the fields as well. You can more or less copy it from the CDDL.

packages/sdk/src/plt/TokenOperation.ts
```diff
@@ -14,6 +14,19 @@ export enum TokenOperationType {
     RemoveDenyList = 'removeDenyList',
     Pause = 'pause',
     Unpause = 'unpause',
+    UpdateMetadata = 'updateMetadata',
+    AssignAdminRoles = 'assignAdminRoles',
+    RevokeAdminRoles = 'revokeAdminRoles',
+}
+
+export enum TokenAdminRole {
```
limemloh, line 26: **Minor thing (non-blocking):** Please add some documentation here

limemloh (approved):

packages/sdk/src/plt/Token.ts
rbaconcordium: changing

packages/sdk/src/plt/TokenOperation.ts
rbaconcordium: adding

packages/sdk/src/plt/TokenOperation.ts
rbaconcordium: adding

PR #660, opened 2026-03-04
Rbc 9/rbac events

rbaconcordium:
## Purpose
Implementing new TokenModuleEvents, TokenUpdateMetadaEvent, TokenAssignAdminRolesEvent, TokenRevokeAdminRolesEvent.

## Changes
new types for the events inside TokenModuleEvent.ts and tests in TokenModuleEvent.ts.test

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/packages/sdk/src/plt/TokenModuleEvent.ts
+++ b/packages/sdk/src/plt/TokenModuleEvent.ts
@@ -1,6 +1,6 @@
 import { EncodedTokenModuleEvent, TransactionEventTag } from '../types.js';
 import { cborDecode } from '../types/cbor.js';
-import { TokenOperationType } from './TokenOperation.js';
+import { TokenOperationType, TokenUpdateAdminRolesDetails } from './TokenOperation.js';
 import { CborAccountAddress, TokenId, TokenMetadataUrl } from './index.js';
 
 type GenTokenModuleEvent<E extends TokenOperationType, T extends Object> = {
@@ -84,6 +84,16 @@ export type TokenUnpauseEvent = GenTokenModuleEvent<TokenOperationType.Unpause, 
 export type TokenUpdateMetadataEvent = GenTokenModuleEvent<TokenOperationType.UpdateMetadata, TokenMetadataUrl.Type>;
 
 /**
+ * An event occuring as the result of an "assignAdminRoles" operation, describing the target of the role assignment.
+ */
+export type TokenAssignAdminRolesEvent = GenTokenModuleEvent<TokenOperationType.AssignAdminRoles, TokenUpdateAdminRolesDetails>;
+
+/**
+ * An event occuring as the result of a "revokeAdminRoles" operation, describing the target of the role revocation.
+ */
+export type TokenRevokeAdminRolesEvent = GenTokenModuleEvent<TokenOperationType.RevokeAdminRoles, TokenUpdateAdminRolesDetails>;
+
+/**
  * A union of all token module events.
  */
 export type TokenModuleEvent =
@@ -93,7 +103,9 @@ export type TokenModuleEvent =
     | TokenRemoveDenyListEvent
     | TokenPauseEvent
     | TokenUnpauseEvent
-    | TokenUpdateMetadataEvent;
+    | TokenUpdateMetadataEvent
+    | TokenAssignAdminRolesEvent
+    | TokenRevokeAdminRolesEvent;
 
 function parseTokenListUpdateEventDetails(decoded: unknown): TokenListUpdateEventDetails {
     if (typeof decoded !== 'object' || decoded === null) {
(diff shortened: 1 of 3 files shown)


packages/sdk/src/plt/TokenModuleEvent.ts
```diff
@@ -78,6 +78,18 @@ export type TokenPauseEvent = GenTokenModuleEvent<TokenOperationType.Pause, Toke
...
+ * An event occuring as the result of an "updateMetadata" operation, describing the URL of the metadata and an optional checksum.
+ */
+export type TokenUpdateMetadataEventDetails = {
+    /** The URL of the metadata. */
+    url: string;
+    /** An optional SHA256 checksum of the metadata contents. */
+    checksum?: string;
+};
+
+export type TokenUpdateMetadataEvent = GenTokenModuleEvent<TokenOperationType.UpdateMetadata, TokenUpdateMetadataEventDetails>;
```
soerenbf: There's already a metadata url type:

```suggestion
export type TokenUpdateMetadataEvent = GenTokenModuleEvent<TokenOperationType.UpdateMetadata, TokenMetadataUrl>;
```

packages/sdk/src/plt/TokenModuleEvent.ts
```diff
@@ -134,6 +155,8 @@ export function parseTokenModuleEvent(event: EncodedTokenModuleEvent): TokenModu
         case TokenOperationType.Pause:
         case TokenOperationType.Unpause:
             return { ...event, type: event.type, details: parseTokenPauseEventDetails(decoded) };
+        case TokenOperationType.UpdateMetadata:
+            return { ...event, type: event.type, details: parseTokenUpdateMetadataEventDetails(decoded) };
```
soerenbf: Consequently, you can 
```suggestion
            return { ...event, type: event.type, details: TokenMetadataUrl.fromCBORValue(decoded) };
```

packages/sdk/test/ci/plt/TokenModuleEvent.test.ts
```diff
@@ -68,6 +69,24 @@ describe('PLT TokenModuleEvent', () => {
         expect(parsedEvent.details).toEqual({});
     });
 
+    it('parses updateMetadata event', () => {
+        const details: TokenUpdateMetadataEventDetails = {
+            url: 'https://example.com/token-metadata.json',
+            checksum: 'abc123',
+        };
```
soerenbf: And this should be changed as well to `TokenMetadataUrl`

packages/sdk/test/ci/plt/TokenModuleEvent.test.ts
soerenbf, line 1: I think it would be good with a test fixture for each event added to verify that the encoding does not change between SDK versions.

packages/sdk/src/plt/TokenModuleEvent.ts
rbaconcordium: changing

packages/sdk/src/plt/TokenModuleEvent.ts
rbaconcordium: changing

packages/sdk/test/ci/plt/TokenModuleEvent.test.ts
rbaconcordium: changing

packages/sdk/test/ci/plt/TokenModuleEvent.test.ts
rbaconcordium: i have added a few tests to parse the structure for the three events, updatemetadata, assign and revoke admin roles.

soerenbf: What I mean is, it would be good to have a fixed CBOR encoding that you test the parsing functionality against. Encoding and subsequently decoding the types is OK for testing that you get the same value out, but to test that nothing breaks between versions you need to test against a fixed encoding as well 🙂

Have a look at how it's done for other types. There are cases of this for other events already in this file.

squirer (approved):

packages/sdk/test/ci/plt/TokenModuleEvent.test.ts
rbaconcordium: added tests for the events using a fixed hex generated by the current cbor encode and using those to test also.

packages/sdk/src/plt/TokenModuleEvent.ts
```diff
@@ -108,6 +132,30 @@ function parseTokenPauseEventDetails(decoded: unknown): TokenPauseEventDetails {
...
+    if (typeof decoded !== 'object' || decoded === null) {
+        throw new Error(`Invalid event details: ${JSON.stringify(decoded)}. Expected an object.`);
+    }
+
+    if (!('account' in decoded && CborAccountAddress.instanceOf(decoded.account))) {
+        throw new Error(`Invalid event details: ${JSON.stringify(decoded)}. Expected 'account'`);
+    }
+
+    return decoded as TokenUpdateAdminRolesDetails;
+}
```
td202: These two functions seem to be identical - why do we have both?

td202: It looks like this duplicates `parseTokenUpdateAdminRolesEventDetails`. We should only have one.

td202 (approved):

packages/sdk/src/plt/TokenModuleEvent.ts
rbaconcordium: ah yes, removing the parseTokenAssign

rbaconcordium: yes, removing this one

PR #661, opened 2026-03-05
Introducing the new proto items to be in the node-js sdk

rbaconcordium:
## Purpose
To introduce the recent additions and changes on the proto files and enabling these to be used within the node sdk.
service proto introduced getTokenAuthorizations
types proto introduced TokenAuthorizationsRequest

## Changes
1. Pointing to new protobuf definitions from concordium base, changed the deps/concordium-base branch to feature/rbac
2. Created inside types.ts, two new types to represent TokenAuthorizationsRequest and TokenAuthorizations
3. in translation.ts, a new function to translate proto TokenAuthorizations to sdk type of TokenAuthorizations
4. in GrpcClient.ts, created a new function, getTokenAuthorizations, given a TokenAuthorizationsRequest
5. fixed external_functions.rs, due to failure in compiling as cbor_encode returns Vec<u8> now and map_err no longer applicable for that cbor_encode.

## Checklist

      hard-to-understand areas.
- [x ] (If necessary) I have updated the CHANGELOG.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/packages/sdk/src/grpc/GRPCClient.ts
+++ b/packages/sdk/src/grpc/GRPCClient.ts
@@ -38,6 +38,7 @@ import type { BlockItemStatus, BlockItemSummary } from '../types/blockItemSummar
 import { countSignatures, isHex, isValidIp, mapRecord, mapStream, unwrap } from '../util.js';
 import * as translate from './translation.js';
 import type { Upward } from './upward.js';
+import { TokenAuthorizationsRequest } from '../grpc-api/v2/concordium/types.js';
 
 /**
  * @hidden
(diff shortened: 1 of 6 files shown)


deps/concordium-base
soerenbf, line 1: Is the intention also to support the new service query in this PR, or is it just to update the reference to `concordium-base`? 🤔

rbaconcordium: yes to support the new service as well, this is just the first commit or so, just to check if this is correct. Making sure all good to start, as mentioned there were some changes also required on the external_functions.rs as cbor_encode return type was changed in base. not going to merge anything yet till after the new query is done

squirer (approved):

packages/sdk/src/grpc/GRPCClient.ts
```diff
@@ -1601,6 +1602,24 @@ export class ConcordiumGRPCClient {
         const tokenIds = this.client.getTokenList(blockHashInput, { abort: abortSignal }).responses;
         return mapStream(tokenIds, PLT.TokenId.fromProto);
     }
+
+    /**
+     * Get the authorizations of a given token in the given block
+     * @param request tokenAuthorizationsRequest structure containing the token ID and block hash
+     * @returns
+     */
+    async getTokenAuthorizations(request: PLT.TokenAuthorizationsRequest): Promise<PLT.TokenAuthorizations> {
```
limemloh: Maybe take the token id and block hash directly instead of through an object, since this is what we do for the other queries.
Also support the optional block hash here.

packages/sdk/src/plt/types.ts
```diff
@@ -74,3 +75,13 @@ export type CreatePLTPayload = {
...
+
+export type TokenAuthorizationsRequest = {
+    blockHash: BlockHash.Type;
+    tokenId: TokenId.Type;
+};
+
+export type TokenAuthorizations = {
+    tokenId: TokenId.Type;
+    details: Cbor.Type;
+};
```
limemloh: These types are exposed directly to the user, so they should be documented.
Ideally linking to the client method which produces these.

limemloh (changes requested):

packages/sdk/src/grpc/GRPCClient.ts
rbaconcordium: changing

packages/sdk/src/plt/types.ts
rbaconcordium: i will remove the TokenAuthorizationsRequest as it is not needed anymore in the function getTokenAuthorizations (changed to the tokenid and blockhash, instead).

will put some info for the TokenAuthorizations type as suggested

limemloh (approved):
LGTM

PR #662, opened 2026-03-10
Revise SDK types to match non-dynamic token module implementation

rbaconcordium:
## Purpose
1. Following the changes on cddl, https://github.com/Concordium/concordium-base/pull/892/changes
we need to remove "additional" from tokenModuleAccountState on the SDK.
2. TokenInitializationParameters has no "additional" on the SDK.
3. TokenModuleState, remove the "additional" from the type structure

## Changes
module.ts, amended the type TokenModuleAccountState.
Cbor.test.ts, there is an existing test involving TokenInitializationParameters which was using customField incorrectly, removed that from the test. 
Also for TokenModuleState test in Cbor.test.ts, amended the test.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/packages/sdk/CHANGELOG.md
+++ b/packages/sdk/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Removed additional field from TokenModuleAccountState type
+- Removed `additional` field from TokenModuleAccountState and TokenModuleState types as they never contained any data
 
 ## 12.0.2
 
--- a/packages/sdk/CHANGELOG.md
+++ b/packages/sdk/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Removed additional field from TokenModuleAccountState type
+
 ## 12.0.2
 
 ### Changed
--- a/packages/sdk/src/plt/module.ts
+++ b/packages/sdk/src/plt/module.ts
@@ -41,18 +41,12 @@ export type TokenModuleState = {
  * managed by the Token Kernel, such as the token identifier and account balance.
  *
  * All fields are optional, and can be omitted if the module implementation does not support them.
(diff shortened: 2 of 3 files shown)


td202 (approved):

packages/sdk/src/plt/module.ts
```diff
@@ -41,18 +41,12 @@ export type TokenModuleState = {
  * managed by the Token Kernel, such as the token identifier and account balance.
  *
  * All fields are optional, and can be omitted if the module implementation does not support them.
- * The structure supports additional fields for future extensibility. Non-standard fields (i.e. any
- * fields that are not defined by a standard, and are specific to the module implementation) may
- * be included, and their tags should be prefixed with an underscore ("_") to distinguish them
- * as such.
  */
 export type TokenModuleAccountState = {
```
allanbrondum, line 40: There are two other types where additional data has also been removed, token module state and initialization parameters.

rbaconcordium: yup, missed TokenModuleState, amended it and repushed on the PR.
The initialization has no additional field on the sdk strangely, so no changes.

packages/sdk/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Removed additional field from TokenModuleAccountState type
```
allanbrondum: Remember the other type in the changelog also. It may also be relevant to state in the changelog, that these fields actually never contained any data, and that is why they are removed.

allanbrondum (approved):

packages/sdk/CHANGELOG.md
rbaconcordium: modifying the changelog

PR #667, opened 2026-04-17
[COR-2368] Fix inability to make a zero-knowledge proof in certain cases

Ivan-Mahda:
## Purpose

Fix such that an attribute can both be used in bullet proofs and revealed at the same time. This was not possible for identity based presentations (actually, it depended on the order of the statements, if the prover would fail

## Changes

Removed single statement checks and validation

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/CHANGELOG.md
+++ b/packages/sdk/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Removed
+
+- In zero knowledge proofs removed limitation of a single statement per attribute
+
 ## 12.0.2
 
 ### Changed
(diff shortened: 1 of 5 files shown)


deps/concordium-base
soerenbf, line 1: how come the concordium-base reference was updated as part of this? 🤔

Ivan-Mahda: I looked at how it was done in other PRs
For example:
https://github.com/Concordium/concordium-node-sdk-js/pull/620/changes
And used command: `git submodule update --remote --merge`
The reason for this is that some of the changes for zkp were made in concordium-base, so it also needed to be updated.

soerenbf (approved):

PR #668, opened 2026-04-30
Add support for meta update transactions

soerenbf:
Closes COR-2388
Ref COR-2317

## Purpose

Adds meta update transaction + lock operations, and events.

## Checklist

      hard-to-understand areas.


packages/sdk/src/plt/cbor-parse.ts
```diff
@@ -0,0 +1,51 @@
...
+        throw new Error(`Invalid transfer details: ${JSON.stringify(details)}. Expected 'amount' to be a TokenAmount`);
+    if (!('recipient' in details) || !CborAccountAddress.instanceOf(details.recipient))
+        throw new Error(
+            `Invalid transfer details: ${JSON.stringify(details)}. Expected 'recipient' to be a TokenHolder`
+        );
+    if ('memo' in details && !(details.memo instanceof Uint8Array || CborMemo.instanceOf(details.memo)))
+        throw new Error(
+            `Invalid transfer details: ${JSON.stringify(details)}. Expected 'memo' to be Uint8Array | CborMemo`
+        );
+    return details as TokenTransfer;
```
limemloh: Could we leave out the casting here, or can't TypeScript differ from the context?
It would be nice to get type errors if we expand the type in the future.
Similarly for the parsers below

packages/sdk/src/plt/LockId.ts
```diff
@@ -0,0 +1,164 @@
...
+
+/** JSON representation of a lock identifier. */
+export type JSON = {
+    /** Account index of the account that created the lock. */
+    accountIndex: bigint;
+    /** Sequence number of the transaction that created the lock. */
+    sequenceNumber: bigint;
+    /** 0-based creation order of the lock within the transaction. */
+    creationOrder: bigint;
+};
```
limemloh, line 16: where do we need the json representation again?
I still think we should come up with a more compact identifier representation instead of this.

packages/sdk/.size-limit.ts
```diff
@@ -4,7 +4,7 @@ module.exports = [
     {
         name: 'GRPC web client tree shaking',
         path: 'lib/esm/index.js',
-        limit: '60 KB',
+        limit: '62 KB',
```
limemloh, line 7: Why did it increase in size? I would expect most of this to be shaken off

limemloh (approved):

packages/sdk/src/plt/cbor-parse.ts
soerenbf: Changed it to avoid casting, we did a similar thing in other places.. I don't know why it could not infer it before, I feel like the statements should enable the compiler to 🤷

packages/sdk/src/plt/LockId.ts
soerenbf: This is just the same as we've done for haskell/rust, and we agreed to just do the simple thing when we discussed it for those libraries.

The JSON representation is needed for events coming from the wallet-proxy

packages/sdk/.size-limit.ts
soerenbf: I don't know.. do we care? I don't think it's that big of a deal.. it's pretty hard to trace these things, and I feel like our time is better spent on other issues 😅

PR #669, opened 2026-05-04
Extend `ConcordiumGRPCClient` with lock queries

soerenbf:
Ref COR-2317
Closes COR-2389

## Purpose

Extend GRPC client with lock queries + add lock-related account data.

## Checklist

      hard-to-understand areas.


packages/sdk/src/plt/cbor-types.ts
```diff
@@ -57,14 +70,48 @@ export type TokenModuleAccountState = {
...
 };
 
 /**
- * These parameters are passed to the token module to initialize the token.
- * The token initialization update will also include the ticker symbol,
- * number of decimals, and a reference to the token module implementation.
+ * A token amount locked for an account, pairing a token identifier with the
+ * amount of that token controlled by the lock.
+ */
+export type LockedTokenAndAmount = {
```
limemloh, line 85: This is identical to type `AccountLockAmount`. Is that intentional?

limemloh (approved):

packages/sdk/src/plt/cbor-types.ts
soerenbf: Nope, `AccountLockAmount` is used in the token account state that already refers to a single token and refers to a `lock` through the defined field in the type. `LockedTokenAndAmount` is used in the lock state for a single lock and refers to a `token` through the defined field in the type.

PR #670, opened 2026-05-04
Add `Lock` client

soerenbf:
Ref COR-2317
Closes COR-2390

## Purpose

Adds a high-level `Lock` client for creating and interacting with locks.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/packages/sdk/src/plt/Lock.ts
+++ b/packages/sdk/src/plt/Lock.ts
@@ -182,16 +182,152 @@ class Lock {
     }
 }
 
+/** Handle for a submitted lock-creation transaction. */
+class LockCreateTransaction {
+    public constructor(
+        private readonly grpc: ConcordiumGRPCClient,
+        public readonly transactionHash: TransactionHash.Type
+    ) {}
+
+    /**
+     * Wait until the transaction is finalized and return the created lock.
+     *
+     * @param finalizationTimeoutSeconds Optional timeout in seconds for waiting for finalization.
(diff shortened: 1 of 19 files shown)


soerenbf (commented):
Tested with a local p11 node running the `feature/p11-locks` branch, and the functionality supported on that branch (create lock, query locks, query lock info) works.

packages/sdk/src/plt/Lock.ts
```diff
@@ -0,0 +1,795 @@
...
+    return true;
+}
+
+/**
+ * Check whether the sender can send funds controlled by the lock.
+ *
+ * @param lock The lock to validate against.
+ * @param sender The sender account to validate.
+ * @param details The send operation details, used to validate the locked amount on the source account.
+ * @returns `true` if the sender can send funds controlled by the lock.
```
limemloh, line 593: I think returning true is a bit misleading, since the signature indicates it will return false if I can't send, but it is actually throwing errors instead.
So I would make it either `ensureCanSend(...)` with return type or stick to `canSend(...): bool` and actually produce false instead of errors here.
Likewise for the function below

limemloh (approved):

packages/sdk/src/plt/Lock.ts
soerenbf: Makes sense, I might go with `validateSend` to align with the corresponding functionality in `Token.ts`.

